PR #117, opened 2014-03-31
Stripe API added

tmcpro:
Stripe API added that gives a basic overview of adding clients, charging, and signing users up for a subscription plan.


MichaelBosch:
Was the subscription billing example ever added back?  I can't find how to do subscription billing.

PR #184, opened 2014-09-05
Fixing dead links

DrkSephy:
This pull request removes a reference to a dead link `http://html5up.net/uploads/previews/6742121165068310.jpg`, as well as fixes a reference display of HTML To Jade converter.

Also, there is one more dead link under the `How do I use Socket.io with Hackathon Starter` section, right before `Mongoose Cheatsheet` section. 

There is a reference to `http://hackathonstarter.herokuapp.com/dashboard`, which leads to `Cannot GET /dashboard` error. I left this link as it was, as I am not sure where you would like to redirect this link. I will leave that up to you, just wanted to bring that up as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -650,7 +650,6 @@ I hope it provides some guidance to you.
 
 ###:rose: Custom HTML and CSS Design 101
 [HTML5 UP](http://html5up.net/) has many beautiful templates that you can download for free.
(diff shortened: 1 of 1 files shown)


raunaqrox:
The Readme still has the dashboard link.

PR #215, opened 2015-01-10
Add fig.yml file for docker-based development stack

kytwb:
I don't have mongodb installed on my local machine as I'm now using a 100% docker-based development stack so I can easily switch projects without having to think about what services are running, where, when etc. Open to comments.

### DIFF
--- a/config/secrets.js
+++ b/config/secrets.js
@@ -18,7 +18,7 @@
 
 module.exports = {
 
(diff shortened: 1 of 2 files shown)


sahat:
There was a similar PR in the past for Vagrant. The same reasoning still applies here:
https://github.com/sahat/hackathon-starter/pull/119

kytwb:
Fair enough.

PR #218, opened 2015-01-16
Typo fix.

phlacin:
Just fixing the first typo I saw.


sahat:
Thanks, but the fix should not contain **x** at all.

PR #220, opened 2015-01-23
Added Lob Api example

jromer94:
I've been seeing lob at a lot of hackathons recently, so I decided to add an API example for it.


sahat:
Thank you. A pull request I'd like to see more often -- that does not deviate from the code style of the project.

PR #231, opened 2015-02-11
Put change directory before removing git remote in README

Falkirks:
`git remote rm origin` needs to be run inside the project directory.


sahat:
Ah, good catch.

PR #232, opened 2015-02-12
Updating Flatly theme to 3.3.2

huffn:
Current version was using variables before they had been declared, and the box-shadow function needed to be included in order to work


sahat:
Thanks!

PR #242, opened 2015-03-21
User model doesn't seem to be used in this file

rstormsf:
Please confirm if I'm right


sahat:
It is no longer used, you are right.

PR #252, opened 2015-03-25
fix typos

josephahn:


sahat:
I probably did not have enough coffee when I wrote the README. ;)

PR #253, opened 2015-03-28
Fix radio button alignment on small screens

alex-dixon:
When viewing /account in a window less than 768px, the radio buttons are positioned off the page:
image

Adding the bootstap column class used on other form fields on the page to the radio buttons moves them onto the page and in line with the rest of the form..


sahat:
:+1:

PR #277, opened 2015-05-25
Remove Ordrx API

xasos:
Recently, Ordrx (formetly Ordrin) shut down support for their API as of April 15, 2015: http://www.ordrx.com/

All instances of the API has been removed (they are non-functioning)


xasos:
Also, I think I broke the live demo (http://hackathonstarter.herokuapp.com) by clicking on the Ordrx API example :(

sahat:
@xasos thanks, and no worries, Heroku automatically restarts the Node.js process if it crashes for whatever reason.

PR #290, opened 2015-06-23
Adapting installation commands of the prerequisites in Fedora Linux.

zwarag:
Changing yum to dnf since yum is deprecated in the current version of Fedora.


sahat:
Ah thanks, it's been a while since I have used Fedora.

PR #292, opened 2015-06-30
git clone should be not have --bare flag.

### LINKED ISSUE #291: git clone should be non-bare
Your getting started section (on the readme) says use "git clone --bare ..." but I think it should be non-bare.

No?

bardiharborow:
Fixes #291.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ The easiest way to get started is to clone the repository:
 
 ```bash
 # Get the latest snapshot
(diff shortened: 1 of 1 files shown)


sahat:
You are right, `--bare` did not have the intended consequences. Someone reported `git push` issues to Heroku after cloning hackathon-starter because it had my remote urls. Thanks for the PR.

Issue #294, opened 2015-07-06
Update Deployment Guide (Help Wanted)


sahat:
Deployment guides have been written over a year ago. Website UI flow has likely changed since then.

I would appreciate if someone could go through IBM Bluemix, Heroku, OpenShift, Microsoft Azure step by step deployment guides in the [README](https://github.com/sahat/hackathon-starter#deployment), update them as necessary, then submit a pull request.

- [x] IBM Bluemix
- [x] Heroku
- [x] OpenShift
- [x] Azure

PR #303, opened 2015-07-31
Updates for one-step-deployment to Heroku

fenwick67:
Updates to the docs for Heroku one-click deploy per #294.

`heroku addons:add` was deprecated and replaced by `heroku addons:create`

I also updated `secrets.js` to automatically pick up the `MONGOLAB_URI` environment var set by the Mongolab add-on in the Heroku environment to make deployment dead-simple.


sahat:
Nice, thanks for updating the README.

PR #308, opened 2015-08-23
Request to merge in bitcore api example

physicsdude:
Hi, please review this request to merge in the awesome bitcore library/api - more info at http://bitcore.io
Thanks!


sahat:
@physicsdude Is there any benefit in having two similar Bitcoin libraries and API examples? What are the differences between Bitcore and BitGo?

physicsdude:
Good question. By the way, I am not associated with either company, though I have met Benny, the author of the BitGo api example for hackathon-starter.

BitGo is really cool, however it is really a multisig wallet service. It's a good way, perhaps, for people to incorporate bitcoin payments or wallet functionality into an app. However, it is one specific company and my understanding of their API is that you cannot do everything that someone who wants to do a hackathon project related to Bitcoin can do.

Essentially, BitGo is a centralized "Wallet API" whereas Bitcore allows you to do pretty much *anything* which it is possible to do with Bitcoin - either using Bitpay's Insight API or your own API based on your own Bitcoin full node. Therefore you can develop a hackathon app using their API and then if this grows into a bigger project or 'real' company, you are not locked into using someone else's api.

The major advantage to including bitcore is that it allows low level bitcoin projects to be created, e.g. creating your own wallet vs using BitGo's, without relying on any third party system. It is also easier to perform lower level operations, such as constructing individual transactions with custom information in the OP_RETURN, handling UTXOs, etc.

With the included libraries, someone can construct the front end for a bitcoin (or other cryptocurrency) related project, and run their own *completely independent* back end 'API' by running a full bitcoin node and running the "Insight" software on that node.

To give a concrete example, at the Money 20/20 hackathon, I worked on something called FreezeRay. This is a custom bitcoin wallet and it is necessary for the wallet to be able to handle the low level operations such as dealing with UTXOs or "Unspent Transaction Outputs" and it also required building the transaction 'step by step'. This would not have been possible with BitGo.

Another thing to consider is that it is possible to build *altcoin* projects using bitcore. This may require hacking bitcore itself, but since you can run your own full altcoin node and have access to all of the bitcore source and can run your own api, you could build altcoin apps which it is not possible to build with BitGo, since BitGo is focused on Bitcoin and you cannot edit their api itself.

Finally, a really cool feature is that it is possible to use the features of bitcore in browser based javascript only, by building it with bower. There are many bitcoin wallets and related things which are purely browser-side js based so that people can download them, inspect the code, and run them offline (for security reasons) - for example, paper wallets. Bitcore allows people to do this whereas something like BitGo wouldn't make sense for this.

Again, I think BitGo is cool but there is a lot you can't do with it, and a lot of people may not want to use their API/wallet service and they may prefer a more comprehensive and open way to work with cryptocurrencies.

p.s.
Here are some examples https://github.com/bitpay/bitcore/blob/master/docs/examples.md

Note that the reason I did this is that the Blockchain Meetup group which I co-organize in Phoenix is hosting a mini hackathon, called a HackAttack. I wanted people to be able to choose between something that supports low level operations, like bitcore, or they can use the BitGo api if they want an easy ready to use multisig wallet api.

physicsdude:
Here's one of the original authors of bitcore talking about bitcore while employed by BitGo :) https://www.youtube.com/watch?v=Qmo3zdYSqxM

physicsdude:
Here's an example of a 'cutting edge' project that's possible with bitcore but not with something like BitGo. A full "Simplified Payment Verification" (SPV) wallet that interacts with the Bitcoin network peer-to-peer through javascript
https://github.com/mappum/webcoin/search?utf8=✓&q=bitcore

sahat:
Thanks for such detailed response. You should have just mentioned that one
is a bitcoin service and another one is a open-source MIT library and I
would have accepted this PR since I don't know whats what in Bitcoin. I
always welcome self-hosted, do-it-yourself solutions.

physicsdude:
OK, thanks. Guess I got a little carried away, obviously this stuff is pretty interesting to me. -Bryan

sahat:
@physicsdude Could you remove client-side version of bitcore from the *public* folder? Even though it is not being referenced and loaded, I like to keep that directory for very common libs that can be used across any number of projects, such as jQuery or Bootstrap.js, whereas having bitcore.js there will only be useful to someone who is building a Bitcore.js-based hack. Keep in mind, this project is quite popular now, so I need to be careful with what gets added and where.

Or am I misunderstanding something and this client-side JS file is required for the Bitcore example to run?

physicsdude:
@sahat you're correct that the client-side version is not required for the nodejs version of bitcore to work. I'll remove the browser side js file and instead put a note in the README about how to build the browser side js in case someone wants to use it.

physicsdude:
Thanks for merging :+1:

sahat:
Thanks @physicsdude for the PR. By the way, is there any particular reason for using `"bitcorePrivateKeyWIF" in req.session` instead of `req.session.bitcorePrivateKeyWIF`?

physicsdude:
If i recall the reason is that req.session.bitcorePrivateKeyWIF can trigger
an error if privateKeyWIF doesnt exist in req.session wheras the other
method will simply return false.

sahat:
@physicsdude I am going to remove Bitcore example from this project for the following reasons:
- Not supported on Windows. https://github.com/bitpay/bitcore-node/issues/271
- Does not support Node 5.0.

There are 2 open issues about this where people are having trouble with installing npm dependencies because of bitcore.

PR #319, opened 2015-09-16
align the styling on /login, /profile and /account

hagerhink:
I felt that the page "jumped around" when navigating between the login, profile and account, so I aligned all of the fields in the forms on the different pages.

screen shot 2015-09-16 at 18 52 49
screen shot 2015-09-16 at 18 52 47
screen shot 2015-09-16 at 18 52 44


sahat:
Thanks!

PR #320, opened 2015-09-18
stripe POST to api/stripe uses 'source' not 'card' in the stripe.charges.create method

reesretuta:
need to update stripe create charge method.  param SOURCE instead of CARD should be used to create a Stripe Charge https://stripe.com/docs/api#create_charge


sahat:
:+1:

PR #346, opened 2015-12-27
Remove multer from README.md list of packages

leoj3n:
According to changelog the multer package is no longer included.


sahat:
:+1:

PR #347, opened 2015-12-29
disabled horizontal resizing of textarea

egladman:
removed horizontal `textarea` resizing as not to break responsiveness when adjusted.


sahat:
Thanks, looks good to me.

PR #370, opened 2016-02-06
Redirection after login

isabasan:


sahat:
@isabasan Could you please provide some context and/or description as to what this PR is supposed to do and why it needs to be changed?

isabasan:
Hello @sahat ,

There was a line to redirect after login but there is nothing to set session "returnTo" variable.

I added that line. Then we need to set "returnTo" null after successfully login. These is a little fixing.

I'm new to github. I reviewed just 3 lines of code but there is 3 commits. It's my fail for first contribution.

P.S. I have to say a big thank you to you for this project. I used an up vote on producthunt too :)

sahat:
Thanks, I appreciate the PR, but this particular case was solved using regex in app.js. It should no longer get stuck in the same returnTo state after multiple log in attempts. Home and contact pages are now also valid returnTo urls.

PR #372, opened 2016-02-08
Contact Form UI Improvement

isabasan:
If user is logged in, "email" and "name" parameters will set for contact form.

### DIFF
--- a/controllers/contact.js
+++ b/controllers/contact.js
@@ -13,7 +13,9 @@ var transporter = nodemailer.createTransport({
  */
 exports.getContact = function(req, res) {
   res.render('contact', {
-    title: 'Contact'
+    title: 'Contact',
+    name: req.user ? req.user.profile.name : '',
+    email: req.user ? req.user.email : ''
   });
 };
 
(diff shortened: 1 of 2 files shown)


sahat:
Unnecessary change.

Also, passing `user` fields to Contact template is redundant since `user` object is already accessible from all templates thanks to the middleware on line 94 of app.js. But even then, you have to use `value=user && user.profile.name` and `value=user && user.email` to set existing values, which is not very clean.

Furthermore, if the *Name* field is set, it will still `autofocus` *Name* field instead of the next unset field like *Email* or *Body*. To handle this case, we will need even more conditional logic: if no name is set, autofocus name field; if name is set, autofocus email; if name and email are set, autofocus body textarea.
screenshot 2016-02-08 13 53 10

To sum it up, I am going to keep it as is - simple and easy to understand. If someone needs to auto-set existing user fields on Contact form they can do it themselves really easily.

isabasan:
Hello @sahat ,

It looks like you are right. I've just want to suggest a fine-tune. But it's ok. It should not be a part of hackathon-starter because if it is necessary, anyone can implement it easy.

Thanks

PR #374, opened 2016-02-11
Add semicolon to test/model.js

dok:
It was bothering me..

### DIFF
--- a/test/models.js
+++ b/test/models.js
@@ -11,7 +11,7 @@ describe('User Model', function() {
     user.save(function(err) {
       if (err) return done(err);
       done();
-    })
+    });
   });
 
   it('should not create a user with the unique email', function(done) {


niallobrien:
Technically it's not a syntax error as Javascript does automatic semicolon insertion.

dok:
That's right, how could I have forgotten? I must have believed for a second since it was showing up in my linter.

Regardless, it's good practice and it will follow the same conventions as the rest of the repo.

niallobrien:
Agreed. :+1:

sahat:
I always put semicolons in my code, if it's not there I simply forgot. Thanks!

sahat:
On a related topic, does anyone have experience with mocking models? Current tests rely on hitting a database (i.e. MongoDB must be running), which kind of defeats the main principle of unit testing. I just didn't know how to do it any better back then.

dok:
Could you clarify a little bit more? From what I'm understanding, model tests should be abstracted so that model-related unit tests should not ping the database. Is that what you are implying?

niallobrien:
https://www.npmjs.com/package/mockgoose might be worth taking a look at.

PR #386, opened 2016-02-24
Fix next being undefined for getReset call

dsavenko:
'next' is undefined for getReset call in user controller. This PR fixes it.


sahat:
Thanks!

PR #398, opened 2016-03-09
Renamed occurrences of Mongolab to mLab to address name change

gabemazerogers:


sahat:
@gabemazerogers Do you know if Heroku's environment variables should stay the same, e.g. `MONGOLAB_URI`? Obviously, it could be called anything, as long as it is referenced correctly, but I know Heroku will use those env variables if available with they 1-click deploy process.

gabemazerogers:
Heroku is still using MONGOLAB_URI for existing applications. I'm not sure about future apps though.

sahat:
Thanks!

PR #403, opened 2016-03-12
do not invoke session on static content request

twilly86:
removing initialization of session if request is to static content to boost performance.

### DIFF
--- a/app.js
+++ b/app.js
@@ -83,7 +83,14 @@ app.use(session({
   })
 }));
 app.use(passport.initialize());
-app.use(passport.session());
+app.use(function(req, res, next){
+  if(req.url.match('img/')||req.url.match('css/')||req.url.match('assets/')||req.url.match('favicon/'))
+    next(); // do not invoke passport on static requests
+  else{
+      passport.session()(req, res, next)
+  }
+});
+
 app.use(flash());
 app.use(function(req, res, next) {
   if (req.path === '/api/upload') {


sahat:
Hi @twilly86, what is the actual performance increase for this particular project? If it's just a few milliseconds, I am not going to add it.

I wish I could get rid of these middlewares, but they are necessary:

```js
app.use(function(req, res, next) {
  if (/api/i.test(req.path)) {
    req.session.returnTo = req.path;
  }
  next();
});

app.use(function(req, res, next) {
  if (req.path === '/api/upload') {
    next();
  } else {
    lusca.csrf()(req, res, next);
  }
});
```

Adding yet another conditional middleware will only serve to increase complexity, and in this particular case, will also require users to update the regex when they add/remove/change static assets directory structure.

sahat:
I am going to close this for now until a cleaner solution exists. Something like run `passport.session()` if current request path is not part of static assets, assuming Express provides a function something like `app.getStaticPath()`, so that way you don't have to check regex for every folder in /public.

On a related note, I am running into a similar issue with my own custom auth system for React, where user is deserialized from database on each request, even if that request is for `favicon.ico`.

niallobrien:
@sahat With sessions being stored in Mongo, I think this PR is quite nice actually, anything to boost performance and reduce cost of your Mongo instance is good in my eyes.

Realistically, I couldn't see people changing the structure of the static assets directories much, and if so, they've probably already committed to quite a bit of work to introduce an asset pipeline etc., so updating this middleware to reflect their changes/preferences won't be much work for them at all. Just my two cents.

PR #407, opened 2016-03-15
mongoose provides an error message in the callback, we should print that

nathanhleung:
Since mongoose provides an error object in the `.on('error')` callback, it would be better to log that than to simply say "check your MongoDB".

### DIFF
--- a/app.js
+++ b/app.js
@@ -49,8 +49,8 @@ var app = express();
  * Connect to MongoDB.
  */
 mongoose.connect(process.env.MONGODB || process.env.MONGOLAB_URI);
-mongoose.connection.on('error', function() {
-  console.log('MongoDB Connection Error. Please make sure that MongoDB is running.');
+mongoose.connection.on('error', function(err) {
+  console.log('MongoDB Connection Error: ' + err);
   process.exit(1);
 });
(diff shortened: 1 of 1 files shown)


sahat:
I am aware that it provides an error object, I could leave it out completely and let it crash with an error stack trace. The problem is neither of those messages are clear. 

```
ECONNREFUSED 127.0.0.1:27017
```

and

```
C:\Users\sahat\Developer\hackathon-starter\node_modules\mongoose\node_modules\mongodb\lib\server.js:236
        process.nextTick(function() { throw err; })
                                      ^
 Error: connect ECONNREFUSED 127.0.0.1:27017
    at Object.exports._errnoException (util.js:890:11)
    at exports._exceptionWithHostPort (util.js:913:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1057:14)
```

More experienced users will realize what's wrong, but someone who is not familiar with Node.js or MongoDB ecosystem will not. It's been a while, but I am pretty sure I added this message 2 years ago because someone opened an issue about this same problem.

nathanhleung:
Ah I see, thanks for the quick response!

PR #409, opened 2016-03-19
Input required attribute

hisener:


sahat:
Aren't those fields already being validated on the server-side?

hisener:
Yes, but why do not use the html5 required attribute before server-side validation?

sahat:
I wanted to highlight the usage of express-validator error messages.
```js
req.assert('email', 'Email is not valid').isEmail();
req.assert('email', 'Email cannot be blank').notEmpty();
```

But I suppose it wouldn't hurt to add it. One suggestion, could you change it to just `required` and also change `autofocus=true` to just `autofocus`.

PR #415, opened 2016-03-21
Fixes #413 Can't set headers after they are sent

### LINKED ISSUE #413: Can't set headers after they are sent
I think this line should have a return otherwise it could cause a 500 error and crash the server.

https://github.com/sahat/hackathon-starter/blob/master/controllers/api.js#L435

KevinBusse:
Added missing return on error.


sahat:
Thanks, good catch.

PR #435, opened 2016-04-23
improve redirect after login

deiucanta:
Instead of checking for explicit routes, this tries to exclude any unwanted routes and store anything else. All of the following 5 conditions have to be met in order to save the current path

1. User must not be logged in (saves some resources)
2. Route must not be `/login` (you don't want to get back to login after login)
3. Route must not be `/signup` (same reason as above)
4. Route must not start with `/auth` (same reason as above)
5. Path does not contain a dot (this usually matches a file and you don't want to be redirected to `favicon.ico`)

### DIFF
--- a/app.js
+++ b/app.js
@@ -95,8 +95,12 @@ app.use(function(req, res, next) {
   next();
 });
 app.use(function(req, res, next) {
-  // After successful login, redirect back to /api, /contact or /
-  if (/(api)|(contact)|(^\/$)/i.test(req.path)) {
+  // After successful login, redirect back to the intended page
+  if (req.user == undefined &&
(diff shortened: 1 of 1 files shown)


ChrisVent:
```javascript
[...] req.path !== '/login' && req.path !== '/signup' [...]
```
@deiucanta the condition never return True, because the 'path' of request object cant be '/login' and the same time '/signup', you should use OR , not AND

vincentbriglia:
since passport is prevalently used, wouldn't https://github.com/jaredhanson/connect-ensure-login be a better option ?

sahat:
Thanks for the pull request and sorry about the delay in merging it.

PR #436, opened 2016-04-23
paypal urls must be taken from env

deiucanta:


sahat:
@deiucanta What's the advantage of moving those values into env? Are `return_url` and `cancel_url` configurable options in the PayPal dashboard, similar to redirect uri in OAuth?

deiucanta:
@sahat Well, you already have those in `.env.example` and you can use localhost for development and the real domain name for production. Also, you need absolute urls - relative urls will have `http://paypal.com` as a base url. As far as I explored their dashboard I didn't see any configurable options for urls.

sahat:
I forgot I had those URLs in .env. In that case it makes sense to use environment variables in the PayPal API example.

Thanks. :+1:

PR #444, opened 2016-04-28
Dots should be allowed in email.

brabidou:
Just my two cents. It's fairly common to have dots in the user's email address and it will appear to the user that they have not signed up since the email is normalized. My pull request allows for dots in the user's email but still removes plus. I can understand removing the plus since that can be "spammy".


sahat:
Thanks, I didn't realize `sanitize` function disallows dots by default.

Minor nitpick: could you add spaces within the braces (`{ code }`)

james-hoegerl:
Using the latest version and dots are still being removed even with this pull request added. What email normalize package are we using? Something must have changed.

PR #446, opened 2016-04-29
Add [Theme Color] Meta Tag in layout.jade

ChrisVent:
Add "Theme color" meta tag in layout.jade

Similar to this:

Theme color meta tag


sahat:
I don't have an Android device to check, is this a universal feature or does it work only with Material Design themes?

ChrisVent:
@sahat Its a Universal Feature, Currently works in my app (fork)

sahat:
@ChrisVent can you see my comment in https://github.com/sahat/hackathon-starter/issues/447 and also remove compiled CSS files from the commit?

PR #453, opened 2016-05-04
ES6 refactor

vanshady:


vanshady:
The app works with no error, but why does it fail the mocha test?

controllers/api.js
```diff
@@ -1,33 +1,40 @@
...
-var twilio;
-var Linkedin;
-var BitGo;
-var clockwork;
-var paypal;
-var lob;
-var ig;
-var Y;
-var request;
+const validator = require('validator');
```
sahat: The only downside of going back to declaring and requiring modules is it results in 2x-2.2x slower start-up time.

sahat: For example on my desktop PC it's a difference between ~516ms and ~1160ms. But maybe that's fine if we assume that most people will remove libraries and require calls for the code they are not planning to use.

vanshady: Should I keep those vars?

sahat: Yeah keep it as is for now. Maybe you can just change `var` to `let`. I will do some more performance testing later to determine the best approach.

vanshady: No problem.

sahat:
Reviewing it right now.

prashcr:
I just refactored `api.js` myself, before I saw this PR.  
  
There's one problem I spotted right away, that I encountered in my own refactoring.

In the web scraping API, this code will not work as you expect.
```
$('.title a[href^="http"], a[href^="https"]').each(() => {
      links.push($(this));
    });
```
If you try logging `this`, you'll see that it now refers to the global `this`.
Previously, `cheerio`'s `each` function sets `this` in the regular callback function to the current element.  
However, with arrow functions, `this` is bound lexically so that has no effect.
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Invoked_through_call_or_apply

You'll need to change this to:
```
$('.title a[href^="http"], a[href^="https"]').each((i, elem) => {
      links.push($(elem));
    });
```

controllers/api.js
```diff
@@ -140,19 +135,16 @@ exports.getFacebook = function(req, res, next) {
...
-
-  request.get('https://news.ycombinator.com/', function(err, request, body) {
-    var $ = cheerio.load(body);
-    var links = [];
-    $('.title a[href^="http"], a[href^="https"]').each(function() {
+exports.getScraping = (req, res, next) => {
+  request.get('https://news.ycombinator.com/', (err, request, body) => {
+    const $ = cheerio.load(body);
+    const links = [];
+    $('.title a[href^="http"], a[href^="https"]').each(() => {
```
prashcr: I just refactored `api.js` myself, before I saw this PR.  
  
There's one problem I spotted right away, that I encountered in my own refactoring.

In the web scraping API, this code will not work as you expect.
```
$('.title a[href^="http"], a[href^="https"]').each(() => {
      links.push($(this));
    });
```
If you try logging `this`, you'll see that it now refers to the global `this`.
Previously, cheerio's `each` function sets `this` in the regular callback function to the current element.  
However, with the arrow callback function, `this` is bound lexically so that has no effect.
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Invoked_through_call_or_apply

You'll need to change this to:
```
$('.title a[href^="http"], a[href^="https"]').each((i, elem) => {
      links.push($(elem));
    });
```

controllers/api.js
```diff
@@ -658,30 +624,28 @@ exports.getInstagram = function(req, res, next) {
...
+        const location = response.query.results.channel.location;
+        const condition = response.query.results.channel.item.condition;
+        done(null, { location, condition });
       });
     }
-  ], function(err, results) {
-    var quotes = results[0];
-    var weather = results[1];
+  ], (err, results) => {
+    const quotes = results[0];
```
prashcr: This could be made simpler and more idiomatic with array destructuring.
```
const [quotes, weather] = result;
```

sahat:
@vanshady Nice, looks good for the most part. Thanks @prashcr for your feedback as well. I merged this PR into `es6` branch. I will give it another look and if I don't find any issues I will merge it into `master`.

controllers/api.js
vanshady: Thanks

sahat:
Found a similar issue with Mongoose middleware where `this` refers to the global scope:

```js
/**
 * Password hash middleware.
 */
userSchema.pre('save', (next) => {
  const user = this;
```

sahat:
After looking through Mongoose issues, it doesn't looks like it passes model as a callback parameter, so the only way to access it is via `this`.

prashcr:
Unfortunately, I think there's no way around it then. Will have to revert that to using `function`.

http://stackoverflow.com/questions/33774472/mongoose-pre-save-is-using-incorrect-this-context

https://github.com/mochajs/mocha/issues/1856

PR #499, opened 2016-06-09
added editorConfig support

mrdivyansh:
We should add .editorconfig file to define consistent coding style.
Refer [here](http://editorconfig.org/)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


sahat:
Hi @mrdivyansh, thanks for the PR but I cannot merge it for the same reason I closed many other pull requests in the spirit of keeping this project really simple and clutter-free.

PR #533, opened 2016-07-22
Remove Venmo

cassidoo:
The Venmo API is no longer available to new users (https://techcrunch.com/2016/02/26/how-not-to-run-a-platform/), this PR removes it from the project.


sahat:
I am hearing first about this. In that case there is no sense to keep this API example around.

sahat:
Thanks. 👍

PR #557, opened 2016-08-09
Added easy clone support for ES5 variation

kamikazechaser:
Fixed cloning instructions in README.md

Users can now easily clone the ES5 variation of the starter pack.


sahat:
👍

PR #566, opened 2016-09-12
Add missing error guards

dskrepps:
My linter noticed some missing error checks in callbacks. Catching these will prevent uncaught exceptions or unexpected application states in edge cases.


sahat:
:+1: I may have omitted it to keep it simple, as those error checks do add code noise. However it is pretty much required to handle thsese type of errors, if someone is planning to run an app in production.

PR #572, opened 2016-09-23
Updated README to reflect code examples correctly

imcodingideas:
I've updated the READ to reflect the changes I see from the code and what's in the documentation. For example `var app = express();` is no longer like it. We're using const for everything in app.js


sahat:
Thanks. :+1:

Issue #546, opened 2016-07-28
NSP (Node Security Project) issues - 7/28/2016


YasharF:
NSP is identifying 9 vulnerabilities in the dependencies.  This Issue is to track and resolve them:

(+) 9 vulnerabilities found

1. Regular Expression Denial of Service
Package:uglify-js
Installed: 2.2.5
Vulnerable: <2.6.0
Patched: >=2.6.0
Path: hackathon-starter@4.0.1 > jade@1.11.0 > transformers@2.1.0 > uglify-js@2.2.5
More Info: https://nodesecurity.io/advisories/48

2. Incorrect Handling of Non-Boolean Comparisons During.
Package:uglify-js
Installed: 2.2.5
Vulnerable: <= 2.4.23
Patched: >= 2.4.24
Path: hackathon-starter@4.0.1 > jade@1.11.0 > transformers@2.1.0 > uglify-js@2.2.5
More Info: https://nodesecurity.io/advisories/39

3. Denial-of-Service Extended Event Loop Blocking
Package:qs
Installed: 0.6.6
Vulnerable: <1.0.0
Patched: >= 1.x
Path: hackathon-starter@4.0.1 > node-linkedin@0.5.3 > request@2.31.0 > qs@0.6.6
More Info: https://nodesecurity.io/advisories/28

4. Denial-of-Service Memory Exhaustion
Package:qs
Installed: 0.6.6
Vulnerable: <1.0.0
Patched: >= 1.x
Path: hackathon-starter@4.0.1 > node-linkedin@0.5.3 > request@2.31.0 > qs@0.6.6
More Info: https://nodesecurity.io/advisories/29

5. Regular Expression Denial of Service
Package:hawk
Installed: 1.0.0
Vulnerable: < 3.1.3  || >= 4.0.0 <4.1.1
Patched: >=3.1.3 < 4.0.0 || >=4.1.1
Path: hackathon-starter@4.0.1 > node-linkedin@0.5.3 > request@2.31.0 > hawk@1.0.0
More Info: https://nodesecurity.io/advisories/77

6. ReDoS via long string of semicolons
Package:tough-cookie
Installed: 0.9.15
Vulnerable: >=0.9.7 <=2.2.2
Patched: >=2.3.0
Path: hackathon-starter@4.0.1 > node-linkedin@0.5.3 > request@2.31.0 > tough-cookie@0.9.15
More Info: https://nodesecurity.io/advisories/130

7. Regular Expression Denial of Service
[...]

PR #575, opened 2016-09-26
Fix redirect issue in the account page

YasharF:
If a user creates an account with an email address, then goes to `/account`, and links a social login page like Google+ at the bottom of the page, immediately after the  callback from Google+ the user will end up at `/` route (default) instead of /account since `req.session.returnTo` isn't set.

This pull request sets `req.session.returnTo` to `/account` if the request is already from an authenticated user and is from `/account` to have the user go back to the account page after a successful linking of a social login account.

While at it, I also put a minor version bump for twit dependency to explicitly call out the non vulnerable version at https://github.com/sahat/hackathon-starter/issues/546 and establish a line in the sand for when all items in that issue are cleared.

### DIFF
--- a/app.js
+++ b/app.js
@@ -105,6 +105,9 @@ app.use(function(req, res, next) {
       !req.path.match(/^\/auth/) &&
       !req.path.match(/\./)) {
     req.session.returnTo = req.path;
+  } else if (req.user &&
+      req.path == '/account') {
+      req.session.returnTo = '/account';
   }
   next();
 });
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hackathon-starter",
-  "version": "4.2.1",
+  "version": "4.2.2",
   "repository": {
     "type": "git",
     "url": "https://github.com/sahat/hackathon-starter.git"
@@ -55,7 +55,7 @@
     "stripe": "^4.11.0",
     "tumblr.js": "^1.1.1",
     "twilio": "^3.3.1-edge",
-    "twit": "^2.2.4",
+    "twit": "^2.2.5",
     "validator": "^5.6.0"
   },
   "devDependencies": {


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hackathon-starter",
-  "version": "4.2.1",
+  "version": "4.2.2",
```
sahat: Could you revert this back to 4.2.1? I generally bump it manually and note it in the README after a few changes have been accumulated. I will bump the package to 4.3 after this PR since there have been some bigger changes since the 4.2.1.

app.js
```diff
@@ -105,6 +105,9 @@ app.use(function(req, res, next) {
       !req.path.match(/^\/auth/) &&
       !req.path.match(/\./)) {
     req.session.returnTo = req.path;
+  } else if (req.user &&
+      req.path == '/account') {
+      req.session.returnTo = '/account';
```
sahat: returnTo could simply be `req.path` here.

sahat:
LGTM after the changes.

app.js
YasharF: would you rather , I fold the condition in the existing if, instead of the else if?
aka:

```
  if ((!req.user &&
      req.path !== '/login' &&
      req.path !== '/signup' &&
      !req.path.match(/^\/auth/) &&
      !req.path.match(/\./)) 
      || (req.user &&
      req.path == '/account')
```

sahat: I think the first approach is more readable, even if it is more lines of code.

YasharF:
It looks like I nuked my github fork and local repo; the twit dependency isn't showing up any issue without a version bump either.  So I'm just going to submit a new PR with just the /account redirect fix.

YasharF:
Closing this PR in favor of https://github.com/sahat/hackathon-starter/pull/582

PR #577, opened 2016-09-29
Fix Deprecation warning of mongoose.Promise 

starcharles:
I noticed deprecation warning below when I run test and app debugging console.

> (node:34180) DeprecationWarning: Mongoose: mpromise (mongoose's default promise library) is deprecated, plug in your own promise library instead: http://mongoosejs.com/docs/promises.html

To fix this warning I override mongoose's internal promise by node.js global promise function.

### DIFF
--- a/app.js
+++ b/app.js
@@ -47,6 +47,7 @@ const app = express();
 /**
  * Connect to MongoDB.
  */
+mongoose.Promise = global.Promise;
 mongoose.connect(process.env.MONGODB_URI || process.env.MONGOLAB_URI);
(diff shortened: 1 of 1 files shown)


sahat:
@starcharles Are you using Node.js 4.x?

starcharles:
@sahat No, I'm using v6.3.0

sahat:
I am not seeing this issue on my machine, but after looking around, seems like many other people are getting the same warning with Mongoose in certain cases. Doesn't hurt to add it. Thanks!

PR #582, opened 2016-10-05
Fix redirect issue in the account page

YasharF:
If a user creates an account with an email address, then goes to `/account`, and links a social login page like Google+ at the bottom of the page, immediately after the  callback from Google+ the user will end up at `/` route (default) instead of /account since `req.session.returnTo` isn't set.

This pull request sets `req.session.returnTo` to `/account` if the request is already from an authenticated user and is from `/account` to have the user go back to the account page after a successful linking of a social login account.


niallobrien:
@YasharF That's a great suggestion and PR. I vote for this to be merged into `master`.

0x62:
+1 on support for merge

sahat:
Thanks @YasharF :+1:

PR #592, opened 2016-10-11
Adds link when developing locally on startup

niallobrien:
Adds a click-able link to the server address in the console (hold CMD on Mac, CTRL on Windows and click the link). Falls back to previous behaviour if the `env` is not set to `development`.

image


sahat:
Thanks for the PR. I've made some slight changes to it, but it still clickable.

PR #593, opened 2016-10-11
﻿Moves routes & updates Readme

niallobrien:
Moves the routes into a dedicated file and updates the Readme to reflect the changes.

The main benefit to this change, is that it's much easier and quicker to locate a `routes.js` file within your project than it is to open the `app.js` file and navigate to the routes section. This PR ensures that the philosophy of the project remains intact and remains easy to use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,10 @@ Table of Contents
 - [Pro Tips](#pro-tips)
 - [FAQ](#faq)
 - [How It Works](#how-it-works-mini-guides)
-- [Mongoose Cheatsheet](#mongoose-cheatsheet)
+- [Cheatsheets](#cheatsheets)
+    - [ES6](#es6-cheatsheet)
+    - [JavaScript Date](#javascript-date-cheatsheet)
+    - [Mongoose Cheatsheet](#mongoose-cheatsheet)
 - [Deployment](#deployment)
 - [Changelog](#changelog)
 - [Contributing](#contributing)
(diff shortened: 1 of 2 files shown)


0x62:
Disagree, I think that abstracting routes into a separate file adds unnecessary complication, without adding much. If we're considering moving routes to a separate file, I'd suggest using something like [express-json-routes](https://www.npmjs.com/package/express-json-routes) to define routes, rather that just exporting routes directly.

niallobrien:
Every web framework on the planet has their routes in a separate file in fairness, so the concept is not that alien. However, introducing yet another middleware to load routes defined as JSON doesn't seem right to me, i.e. Which of the two is the smallest change from what we have at the moment?  Plus, keeping everything in JS makes it easier for beginners imo.

config/routes.js
```diff
@@ -0,0 +1,111 @@
...
+  });
+  app.get('/auth/steam', passport.authorize('openid', { state: 'SOME STATE' }));
+  app.get('/auth/steam/callback', passport.authorize('openid', { failureRedirect: '/login' }), (req, res) => {
+    res.redirect(req.session.returnTo || '/');
+  });
+  app.get('/auth/pinterest', passport.authorize('pinterest', { scope: 'read_public write_public' }));
+  app.get('/auth/pinterest/callback', passport.authorize('pinterest', { failureRedirect: '/login' }), (req, res) => {
+    res.redirect('/api/pinterest');
+  });
+}
```
lostatseajoshua: Missing a newline at EOF

niallobrien:
Closing as people don't seem that interested.

PR #602, opened 2016-10-20
Adds yarn lock file

niallobrien:
Adds a default yarn lock file for those that don't want to use npm, who want quicker and more deterministic deployments/installs of dependencies etc.


sahat:
@niallobrien In the future, what would be the best way to update yarn.lock for me with new package versions after running npm-check-updates?

niallobrien:
@sahat With `yarn` you can do `yarn upgrade` to update all dependencies. It will also update your `package.json` file and your `yarn.lock` file. See https://yarnpkg.com/en/docs/cli/upgrade
Also, you could do the following to update specific dependencies:
```
yarn upgrade [package]
yarn upgrade [package]@[version]
yarn upgrade [package]@[dist-tag]
```
For further info, see: https://yarnpkg.com/en/docs/managing-dependencies#toc-upgrading-a-dependency

The main advantage to all of this other than speed (`yarn` can install packages that have already been cached locally, which mean you can even install packages offline), is consistent installation of specific dependencies and their versions, which is quite neat.

If you'd prefer, I could add such information to the `Readme.md` and update this PR?

sahat:
Yes please do, I would really appreciate that. Yarn is pretty well known in the the Node.js community now, but might strange for those who are new to Node.js or have been out of the loop for the past week. I've read about the project announcement, its positive benefits, but have yet to give it a try.

niallobrien:
@sahat Sure, no problem. I will also clarify that `yarn upgrade` upgrades packages to the latest version conforming to the version rules set in `package.json`, instead of the exact versions defined in `yarn.lock` and will supply some extra reading material (links).

niallobrien:
@sahat Can you please review the Readme changes I've add to this PR? Thanks in advance.

sahat:
Looks good, thank you @niallobrien! 👍

PR #617, opened 2016-11-04
- added gsd kit by creative tim as a theme.

conacelelena:


sahat:
Thanks @conacelelena  👍!

Issue #622, opened 2016-11-17
Pug/Jade  variable in attribute no longer supported


leesander1:
I ran into this when trying to get the linkedin view to display img correctly.
You can read more about it [here](https://pugjs.org/language/attributes.html).
To fix this issue
`          img.thumbnail(src='#{profile.pictureUrl}')`
can be replaced with
`          img.thumbnail(src='' + profile.pictureUrl)`
I think this occurs in several of the views other than the linkedin pug.

YasharF:
@leesander1 , do you think if you can submit a pull request for this?  I did a quick check and these are the places that I found what you mentioned:

https://github.com/sahat/hackathon-starter/blob/b96713c62bd825015ad826d36e7d5c4804621d7b/views/api/foursquare.pug#L37

https://github.com/sahat/hackathon-starter/blob/b96713c62bd825015ad826d36e7d5c4804621d7b/views/api/github.pug#L28

https://github.com/sahat/hackathon-starter/blob/b96713c62bd825015ad826d36e7d5c4804621d7b/views/api/instagram.pug#L33  five other lines

https://github.com/sahat/hackathon-starter/blob/b96713c62bd825015ad826d36e7d5c4804621d7b/views/api/lastfm.pug and four other lines

https://github.com/sahat/hackathon-starter/blob/b96713c62bd825015ad826d36e7d5c4804621d7b/views/api/linkedin.pug#L26

https://github.com/sahat/hackathon-starter/blob/b96713c62bd825015ad826d36e7d5c4804621d7b/views/api/steam.pug#L22

https://github.com/sahat/hackathon-starter/blob/b96713c62bd825015ad826d36e7d5c4804621d7b/views/api/tumblr.pug#L26

https://github.com/sahat/hackathon-starter/blob/b96713c62bd825015ad826d36e7d5c4804621d7b/views/api/twitter.pug#L48

leesander1:
@YasharF yes I will submit a pull request with the fixes when I have some time tomorrow

YasharF:
The related PR has been merged; closing this issue.

PR #628, opened 2016-12-07
Update pug imgs

leesander1:
This should fix all the jade/pug view changes in https://github.com/sahat/hackathon-starter/issues/622

### DIFF
--- a/views/api/foursquare.pug
+++ b/views/api/foursquare.pug
@@ -5,7 +5,7 @@ block content
     h2
       i.fa.fa-foursquare
       | Foursquare API
-  
+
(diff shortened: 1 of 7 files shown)


sahat:
@leesander1 Sorry for getting back to you so late. Would you mind replacing it with ES6 template strings? I believe Pug supports that at least.

E.g.

```js
img(src=`${user.profile_picture}`, width='75', height='75')
```

kamiranoff:
I don't think you need the `'' +` in the src attribute variables.
That works too : `img.thumbnail(src=profile.pictureUrl)` (Linkedin example).

Issue #625, opened 2016-11-24
Nodemailer/Sendgrid on contact form not working by default on Heroku


nk8:
For setting up on Heroku the contact form mailer doesn't work by default. 

In addition to the Mongodb add-on, the sendgrid addon should also be added.

I simply added by doing `heroku addons:create sendgrid:starter`

And then I had to update user on contact.js controller from `process.env.SENDGRID_USER` to `process.env.SENDGRID_USERNAME`

It might be helpful for others to add these notes to the README?

YasharF:
I was able to get it to work on Heroku without the sendgrid addon in Heroku.  The developer still does need to create an account with Sendgrid.  Basically, I just did step 1 and 2 from the readme.md, and then added the environment variables the the Heroku instance (didn't need to change contact.js)

_From the readme.md:_
1. Go to https://sendgrid.com/user/signup
2. Sign up and confirm your account via the activation email
3. Then enter your SendGrid Username and Password into .env file

So instead of #3, I added the configs through the bash/command prompt via the Heroku Toolbelt/CLI by the command `heroku config:set  SENDGRID_USER=[sendgridUserID] SENDGRID_PASSWORD=[sendgridPassword]` .  Alternatively you can use the Environment Variables under the Settings UI tab of your Heroku instance to add the SENDGRID_USER and SENDGRID_PASSWORD

Same with Mongodb, you don't have to add a Heroku add-on, if you have an instance running somewhere already (even at mlabs), you just need to add the environment variables to your Heroku dyno instance :)

PR #629, opened 2016-12-14
Update to current SendGrid heroku env name

strathausen:
Just tried and Heroku sets it now up with SENDGRID_USERNAME

### DIFF
--- a/controllers/contact.js
+++ b/controllers/contact.js
@@ -3,7 +3,7 @@ const nodemailer = require('nodemailer');
 const transporter = nodemailer.createTransport({
   service: 'SendGrid',
   auth: {
-    user: process.env.SENDGRID_USER,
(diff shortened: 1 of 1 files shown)


YasharF:
I could be wrong, but I don't think if this change is needed if you are following the steps in the readme on setting up Sendgrid.  See my comment at https://github.com/sahat/hackathon-starter/issues/625

I think you may have used the Heroku add-in workflow for adding SendGrid instead of the steps outlined in readme.md.

_From the readme.md:_
1. Go to https://sendgrid.com/user/signup
2. Sign up and confirm your account via the activation email
3. Then enter your SendGrid Username and Password into .env file

strathausen:
Ah I see, okay good point! Thanks for taking the time!

PR #632, opened 2017-01-05
Adapted fb request scope to current definition

RobTS:
Adapted fb request scope to match current scheme as previous one caused an error message (not all fields approved for app).

Properties requested are all covered by new definition: https://developers.facebook.com/docs/facebook-login/permissions/#reference-public_profile

Functionality is the same as before from what I can see.

Screenshot of error: 
screen shot 2017-01-05 at 13 02 45


sahat:
Thanks :+1:

PR #638, opened 2017-01-17
Remove async package and use built-in ES6 Promises

fmcarvalho:
Remove useless `async` package from `hackathon-starter` and replace it by built-in ES6 Promises, `then()` and `all()`, which are equivalent to `async::waterfall()` and `async::parallel()` use.

Side improvements: 
* `createRandomToken()` and `User.findOne(...)` are dispatched in parallel instead of sequentially (as happened before with `waterfall()`) because there is no dependency between each other;
* Suppresses the `async` idiomatic `if (err) { return done(err); }`, because "_If an error is thrown in the executor function, the promise is rejected._" (from [developer.mozilla.org Promise](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise))
* Replaces the idiomatic `(err, results) => { if (err) { return next(err); } ... /* else block */` with `.catch(next);`

### DIFF
--- a/controllers/api.js
+++ b/controllers/api.js
@@ -1,6 +1,7 @@
 'use strict';
 
-const request = require('request');
+const bluebird = require('bluebird');
+const request = bluebird.promisifyAll(require('request'), { multiArgs: true });
 const cheerio = require('cheerio');
 const graph = require('fbgraph');
 const LastFmNode = require('lastfm').LastFmNode;
@@ -13,7 +14,7 @@ const Linkedin = require('node-linkedin')(process.env.LINKEDIN_ID, process.env.L
 const clockwork = require('clockwork')({ key: process.env.CLOCKWORK_KEY });
 const paypal = require('paypal-rest-sdk');
 const lob = require('lob')(process.env.LOB_KEY);
-const ig = require('instagram-node').instagram();
+const ig = bluebird.promisifyAll(require('instagram-node').instagram());
 const foursquare = require('node-foursquare')({
   secrets: {
     clientId: process.env.FOURSQUARE_ID,
@@ -22,25 +23,8 @@ const foursquare = require('node-foursquare')({
   }
 });
 
-/**
- * Returns a function that will wrap the given func.
- * Instead of taking a callback, the returned function will
- * return a promise whose fate is decided by the callback
- * behavior of the given node function.
- */
-function promisify(func) {
-  return function funcWrapper(...funcArgs) {
-    const target = this;
-    return new Promise((resolve, reject) => {
-      funcArgs.push((err, ...data) => {
-        if (err) return reject(err);
-        if (data.length === 1) resolve(data[0]);
-        else resolve(data);
-      });
-      func.apply(target, funcArgs);
-    });
-  };
-}
+foursquare.Venues = bluebird.promisifyAll(foursquare.Venues);
+foursquare.Users = bluebird.promisifyAll(foursquare.Users);
 
 /**
  * GET /api
@@ -59,9 +43,9 @@ exports.getApi = (req, res) => {
 exports.getFoursquare = (req, res, next) => {
   const token = req.user.tokens.find(token => token.kind === 'foursquare');
   Promise.all([
-    promisify(foursquare.Venues.getTrending)('40.7222756', '-74.0022724', { limit: 50 }, token.accessToken),
-    promisify(foursquare.Venues.getVenue)('49da74aef964a5208b5e1fe3', token.accessToken),
-    promisify(foursquare.Users.getCheckins)('self', null, token.accessToken)
+    foursquare.Venues.getTrendingAsync('40.7222756', '-74.0022724', { limit: 50 }, token.accessToken),
+    foursquare.Venues.getVenueAsync('49da74aef964a5208b5e1fe3', token.accessToken),
+    foursquare.Users.getCheckinsAsync('self', null, token.accessToken)
   ])
   .then(([trendingVenues, venueDetail, userCheckins]) => {
     res.render('api/foursquare', {
@@ -187,33 +171,30 @@ exports.getLastfm = (req, res, next) => {
     api_key: process.env.LASTFM_KEY,
     secret: process.env.LASTFM_SECRET
   });
-  Promise.all([
-    (function artistInfo() {
-      return new Promise((resolve, reject) => {
-        lastfm.request('artist.getInfo', {
-          artist: 'Roniit',
-          handlers: {
-            success: resolve,
-            error: reject
-          }
-        });
+  const artistInfo = () =>
+    new Promise((resolve, reject) => {
+      lastfm.request('artist.getInfo', {
+        artist: 'Roniit',
+        handlers: {
+          success: resolve,
+          error: reject
+        }
       });
-    }()),
-    (function artistTopTracks() {
-      return new Promise((resolve, reject) => {
-        lastfm.request('artist.getTopTracks', {
-          artist: 'Roniit',
-          handlers: {
-            success: (data) => {
-              resolve(data.toptracks.track.slice(0, 10));
-            },
-            error: reject
-          }
-        });
+    });
+  const artistTopTracks = () =>
+    new Promise((resolve, reject) => {
+      lastfm.request('artist.getTopTracks', {
+        artist: 'Roniit',
+        handlers: {
+          success: (data) => {
+            resolve(data.toptracks.track.slice(0, 10));
+          },
+          error: reject
+        }
       });
-    }()),
-    (function artistTopAlbums() {
-      return new Promise((resolve, reject) => {
+    });
+  const artistTopAlbums = () =>
+      new Promise((resolve, reject) => {
         lastfm.request('artist.getTopAlbums', {
           artist: 'Roniit',
           handlers: {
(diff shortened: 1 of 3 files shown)


controllers/api.js
```diff
@@ -24,6 +23,26 @@ const foursquare = require('node-foursquare')({
 });
 
 /**
+ * Returns a function that will wrap the given func.
+ * Instead of taking a callback, the returned function will
+ * return a promise whose fate is decided by the callback
+ * behavior of the given node function.
+ */
+function promisify(func) {
```
sahat: I'd prefer if we could use bluebird's promisify function. See example here: https://github.com/sahat/megaboilerplate/blob/master/generators/utils.js#L9

controllers/api.js
```diff
@@ -39,32 +58,20 @@ exports.getApi = (req, res) => {
...
-    },
-    userCheckins: (callback) => {
-      foursquare.Users.getCheckins('self', null, token.accessToken, (err, results) => {
-        callback(err, results);
-      });
-    }
-  },
-  (err, results) => {
-    if (err) { return next(err); }
+  Promise.all([
```
sahat:
```js
const getTrending = Promise.promisify(foursquare.Venues.getTrending);

Promise.all([
  getTrending)('40.7222756', '-74.0022724', { limit: 50 }, token.accessToken)
  ...
  ...
]);
```

It may be even possible to promisifyAll the entire fourssquare library at the very top. In the example linked above, I haven't had much luck with `fs` module so I only promisified functions that were used in there.

controllers/api.js
```diff
@@ -180,64 +187,62 @@ exports.getLastfm = (req, res, next) => {
...
-      lastfm.request('artist.getInfo', {
-        artist: 'Roniit',
-        handlers: {
-          success: (data) => {
-            done(null, data);
-          },
-          error: (err) => {
-            done(err);
+  Promise.all([
+    (function artistInfo() {
```
sahat: Would prefer to have function expressions declared outside:

```js
const artistInfo = () => {
  return new Promise((resolve, reject) => {...}
};

File: controllers/api.js [hunk 4]
```diff
@@ -180,64 +187,62 @@ exports.getLastfm = (req, res, next) => {
...
-      lastfm.request('artist.getTopTracks', {
-        artist: 'Roniit',
-        handlers: {
-          success: (data) => {
-            done(null, data.toptracks.track.slice(0, 10));
-          },
-          error: (err) => {
-            done(err);
+    }()),
+    (function artistTopTracks() {
```
sahat: Same as above.

controllers/api.js
```diff
@@ -180,64 +187,62 @@ exports.getLastfm = (req, res, next) => {
...
-      lastfm.request('artist.getTopAlbums', {
-        artist: 'Roniit',
-        handlers: {
-          success: (data) => {
-            done(null, data.topalbums.album.slice(0, 3));
-          },
-          error: (err) => {
-            done(err);
+    }()),
+    (function artistTopAlbums() {
```
sahat: Same as above.

controllers/api.js
```diff
@@ -296,45 +301,49 @@ exports.postTwitter = (req, res, next) => {
 exports.getSteam = (req, res, next) => {
   const steamId = '76561197982488301';
   const params = { l: 'english', steamid: steamId, key: process.env.STEAM_KEY };
-  async.parallel({
-    playerAchievements: (done) => {
+  const fetch = promisify(request.get);
```
sahat: Use bluebird's promisify.

controllers/api.js
```diff
@@ -296,45 +301,49 @@ exports.postTwitter = (req, res, next) => {
...
+  Promise.all([
+    (function playerAchievements() {
```
sahat: Function expressions here as well.

controllers/api.js
```diff
@@ -458,37 +467,22 @@ exports.getInstagram = (req, res, next) => {
...
-    },
-    myRecentMedia: (done) => {
-      ig.user_self_media_recent((err, medias) => {
-        done(err, medias);
-      });
-    }
-  }, (err, results) => {
-    if (err) { return next(err); }
+  Promise.all([
+    promisify(ig.user_search)('richellemead'),
```
sahat: Similar to Foursquare's comments, try using bluebird's promisifyAll on the entire `ig` module.

controllers/user.js
```diff
@@ -250,51 +249,52 @@ exports.postReset = (req, res, next) => {
...
-          });
-        });
-    },
-    function sendResetPasswordEmail(user, done) {
-      const transporter = nodemailer.createTransport({
-        service: 'SendGrid',
-        auth: {
-          user: process.env.SENDGRID_USER,
-          pass: process.env.SENDGRID_PASSWORD
+  function resetPassword() {
```
sahat: Please use function expressions for inner functions below, e.g. `const resetPassword = () { }`.

controllers/user.js
```diff
@@ -250,51 +249,52 @@ exports.postReset = (req, res, next) => {
...
-        subject: 'Your Hackathon Starter password has been changed',
-        text: `Hello,\n\nThis is a confirmation that the password for your account ${user.email} has just been changed.\n`
-      };
-      transporter.sendMail(mailOptions, (err) => {
-        req.flash('success', { msg: 'Success! Your password has been changed.' });
-        done(err);
+  }
+
+  function sendResetPasswordEmail(user) {
+    if (!user) return;
```
sahat: Missing braces around return.

controllers/user.js
```diff
@@ -250,51 +249,52 @@ exports.postReset = (req, res, next) => {
...
+        if (!user) {
+          req.flash('errors', { msg: 'Password reset token is invalid or has expired.' });
+          return res.redirect('back');
         }
+        user.password = req.body.password;
+        user.passwordResetToken = undefined;
+        user.passwordResetExpires = undefined;
+        return user.save().then(() => new Promise((resolve, reject) => {
+          req.logIn(user, (err) => {
+            if (err) return reject(err);
```
sahat: Missing braces around return.

controllers/user.js
```diff
@@ -325,51 +325,51 @@ exports.postForgot = (req, res, next) => {
...
-  });
+  }
+  return Promise.all([
+    createRandomToken(),
+    User.findOne({ email: req.body.email }).exec()
+  ])
+  .then(setRandomToken)
+  .then(sendForgotPasswordEmail)
+  .then(() => res.redirect('/forgot'))
+  .catch(err => next(err));
```
sahat: I believe this could simply be `catch(next)`.

controllers/user.js
```diff
@@ -325,51 +325,51 @@ exports.postForgot = (req, res, next) => {
...
       });
-    }
-  ], (err) => {
-    if (err) { return next(err); }
-    res.redirect('/forgot');
-  });
+  }
+  return Promise.all([
+    createRandomToken(),
+    User.findOne({ email: req.body.email }).exec()
```
sahat: I understand the reason why you split User.findOne from setRandomToken, but I'd prefer to have it back there to make this promise flow simpler. Then we  don't even need to use Promise.all here.

controllers/api.js
fmcarvalho: Done. I removed it and replace it by bluebird.

controllers/api.js
fmcarvalho: DONE. I used promisifyAll. Yet, int this case I did it in two steps because foursquare contains inner objects.

```js
foursquare.Venues = bluebird.promisifyAll(foursquare.Venues);
foursquare.Users = bluebird.promisifyAll(foursquare.Users);
```

controllers/api.js
fmcarvalho: DONE

controllers/api.js
fmcarvalho: DONE

controllers/api.js
fmcarvalho: DONE

controllers/api.js
fmcarvalho: DONE. I removed auxiliary `fetch` variable and promisified whole `request` library:

```js
const request = bluebird.promisifyAll(require('request'), { multiArgs: true });
```

controllers/api.js
fmcarvalho: DONE

controllers/api.js
fmcarvalho: DONE

```js
const ig = bluebird.promisifyAll(require('instagram-node').instagram());
```

controllers/user.js
fmcarvalho: DONE

controllers/user.js
fmcarvalho: FIXED

controllers/user.js
fmcarvalho: FIXED

controllers/user.js
fmcarvalho: DONE. I keep it the original flow.

controllers/user.js
fmcarvalho: FIXED

fmcarvalho:
@sahat I fixed all change requests according to your suggestions.  Thanks

sahat:
Thanks, sorry for taking so long to get back to this pull request.

PR #645, opened 2017-02-06
fix owner is undefined

Clement83:
With user I have an error: `Bad Request Empty value for parameter 'owner': undefined`

It work great with owner :)


sahat:
Thanks.

PR #650, opened 2017-02-10
change year from 2016 to 2017

aamirsahmad:
change copyright year from 2016 to 2017 in views/partials/footer.pug


aamirsahmad:
change copyright year from 2016 to 2017 in views/partials/footer.pug

sahat:
👍

PR #664, opened 2017-03-06
[refactor] - aggregated middleware.

TyreseW:
"You can also load a series of middleware functions together, which creates a sub-stack of the middleware system at a mount point." - ExpressJS documentation ( http://expressjs.com/en/guide/using-middleware.html ).


sahat:
Thank you for the PR @TyreseW, but I am going have to politely decline as I don't see much improvement with these changes in my opinion.

PR #665, opened 2017-03-07
[fix] Heroku re-deploy issue with node-sass

gballet:
When deploying over an already-deployed app in heroku, binaries in the `vendor` directory of module `node-sass` disappear.

This extra post-installation step will force a fetch, for the application to start properly.


sahat:
👍

PR #707, opened 2017-06-05
fixed markdown issue and promises .then() issue

User1m:
Added spacing to properly markdown Mongoose Cheatsheet. Also added missing closing brace from Promises .then() example.


User1m:
@sahat

PR #735, opened 2017-09-26
Update User.js

tutv:


koelle25:
Why do you think this change is needed? For me it breaks the password change function (after changing the password with your PR, there's the plaintext password instead of the hash).

PR #736, opened 2017-10-02
Break out large app.js file into separate smaller files based on functionality

esthersweon:


niallobrien:
I personally don't like how you've split the routes into different files which are hard to identify as files that contain routes. I'd be interested to hear what others think of this.

YasharF:
I really appreciate the effort in refactoring, however the longer app.js with all of the routes in place were implemented by design to favor simplicity.  You can find out more about the rationale behind this design decision at:
https://github.com/sahat/hackathon-starter#why-do-you-have-all-routes-defined-in-appjs

PR #737, opened 2017-10-05
Improvement/eslint

yusufsyaifudin:
This add eslint with airbnb rules and can be check using `gulp lint` command. I've check `npm test` running without errors to ensure that code style does not affect app's logic.


app.js
```diff
@@ -49,7 +51,9 @@ const app = express();
  * Connect to MongoDB.
  */
 mongoose.Promise = global.Promise;
-mongoose.connect(process.env.MONGODB_URI || process.env.MONGOLAB_URI);
+mongoose.connect(process.env.MONGODB_URI || process.env.MONGOLAB_URI, {
```
niallobrien: This should maybe be in a different PR as it's not related to ESlint.

YasharF:
This PR seems to be combining more than just adding eslint, and at this point there are some merge conflict issues as well.  It might be better if this is closed and separate PRs are created for each specific issue.  I would also suggest creating github issues for the improvements that you are thinking about first for discussion.  In that way, other contributors or folks that are active here can comment and reduce the chances of you having to redo PRs, etc.

YasharF:
Closing the PR since there hasn't been an update from the requester and there are also multiple merge conflicts at this point.  There are larger upcoming changes in the project in regards to switching to async/await and finishing ES6 migration which are further going to cause replication effort with this PR.  Perhaps code style and formatting can be discussed to be integrated with the rewrites.

PR #738, opened 2017-10-08
Fix a mongoose deprecation warning

zhang-z:
With latest mongoose, `node app.js` throws warning:
```
DeprecationWarning: `open()` is deprecated in mongoose >= 4.11.0, use `openUri()` instead, or set the `useMongoClient` option if using `connect()` or `createConnection()`. See http://mongoosejs.com/docs/connections.html#use-mongo-client
```
This patch fixes it.

### DIFF
--- a/app.js
+++ b/app.js
@@ -49,7 +49,7 @@ const app = express();
  * Connect to MongoDB.
  */
 mongoose.Promise = global.Promise;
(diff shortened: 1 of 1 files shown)


YasharF:
Thanks for the PR; this is no longer needed because of an update in connect-mongo.  See https://github.com/sahat/hackathon-starter/issues/739#issuecomment-374114738

PR #740, opened 2017-10-09
Fix deprecation warning message in Mongoose and connect-mongo.

### LINKED ISSUE #739: DeprecationWarning: ensureIndex() is deprecated in Mongoose >= 4.12.0, use createIndex() instead
This warning message only appears when I start my new project.
It's probably **connect-mongo** is calling ensureIndex() somewhere.

aquartier:
This PR will resolve #739

### DIFF
--- a/app.js
+++ b/app.js
@@ -49,7 +49,7 @@ const app = express();
  * Connect to MongoDB.
  */
 mongoose.Promise = global.Promise;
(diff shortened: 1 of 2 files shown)


imcodingideas:
This does fix the issue. Thumbs up.

YasharF:
Thank you again for the quick response and the PR submission.  It looks like the problem was because of an issue with connect-mongo which got fixed in its 2.0 release.  I put additional details in the related github issue.

Thanks again.

PR #743, opened 2017-10-11
Update Facebook instructions 

diogeneshamilton:
Made it just a _little_ less difficult to get started with Facebook.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ Features
 Prerequisites
 -------------
 
-- [MongoDB](https://www.mongodb.org/downloads)
+- [MongoDB](https://www.mongodb.org/downloads) needs to be installed and running.
 - [Node.js 6.0+](http://nodejs.org)
 - Command Line Tools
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,7 +101,7 @@ Features
 Prerequisites
 -------------
 
-- [MongoDB](https://www.mongodb.org/downloads)
+- [MongoDB](https://www.mongodb.org/downloads) needs to be installed and running.
```
YasharF: The user doesn't have to download, install, and run MongoDB, since they can use an mLab's hosted database. 

The Readme sections: "Tip" under FAQ -> "I am getting MongoDB Connection Error, how do I fix it?"  and the section about using MongoLab.

YasharF:
If you like to remove the MongoDB stuff instead; we can then just go ahead and merge the Facebook updates without the MongoDB note in your PR.  If so, can you update your PR?

Thanks

README.md
diogeneshamilton: That's fair!

YasharF:
Thank you for the PR :)

PR #757, opened 2017-11-09
Fix google map not showing error

whmsysu:
Fix google map not showing error and add google map api key field to config file


YasharF:
Thank you

PR #759, opened 2017-11-19
OAuth callback URL should be environment-specific variables

git-issue:
###  1. If production project supports https, It would be cause a redirect url  mismatched problem, this would make user confused and spend time to debug，as shown in the following figures.

----
image

---
   image

----
   image

### 2. I think the OAuth callback url, clientID, clientSecret and so on are all configurable. They may be changed in different environment. They should be  keep consistent and  save in the .env files.


YasharF:
Let's discuss this further in the related issue.

YasharF:
Closing this issue as there hasn't been a response from the requester, and considering that the ports in placeholders callback URLs in the PR are not valid.

Issue #725, opened 2017-08-20
Cannot GET /api/bitgo


shahzeb1:
URL: https://hackathonstarter-sahat.rhcloud.com/api/bitgo

Screenshot: 
![image](https://user-images.githubusercontent.com/1383831/29492858-a2900930-853e-11e7-906f-6c389b76962a.png)

JonLim:
Looks like BitGo was removed because of a dependency issue, which was committed in https://github.com/sahat/hackathon-starter/commit/fc464d8728b862e567a4df51c3a4e0747be353bf, and brought up in issue #537.

Might be a good idea to clean it up and remove references to it from the `API Examples` page?

I can take that on, if it's cool?

YasharF:
PR merged.  Thank you.

PR #769, opened 2017-12-16
Cleaning up api/index view to remove reference to BitGo

JonLim:
Per the changelog from July 23, 2016, BitGo was removed due to a dependency issue
with `secp256k1` and Windows. This PR removes the reference on the `API Examples`
page template, addressing issue #725.


YasharF:
Thank you

Issue #770, opened 2017-12-16
Last.fm API Example contains broken images + incorrect data references


JonLim:
It looks like the Last.fm API Example page has broken images, due to a deprecation of interpolation syntax with Pug/Jade, and is also listing the incorrect data for albums and tracks of the artist.

![screen shot 2017-12-16 at 4 43 57 pm](https://user-images.githubusercontent.com/82186/34074600-938b3fc8-e280-11e7-845d-3996392c53a1.png)

I will submit a PR with the fix!

YasharF:
PR merged, closing the issue.  Thank you 👍

PR #771, opened 2017-12-16
Fixing Pug template for Last.fm API Example

JonLim:
Many of the URLs for images and links in the API Example for Last.fm
were incorrectly referenced, due to deprecations with attribute
interpolation syntax.

In addition, the order or promise resolution had mixed up the list
of top tracks and top albums, causing the same image to be shown for
all "top albums", and the list of "top tracks" were actually album
names. This has been corrected.

Addresses issue #770.


YasharF:
Thank you!

PR #773, opened 2018-01-07
Fix Docker files bugs

aptarmy:
- `Dockerfile` needs to expose port `8080` as the default port hard-coded in `app.js`  is `8080`
- `docker-compose.yml` needs to conform `Dockerfile`. If `Dockerfile` and `app.js` expose port `8080` then `docker-compose.yml` must map container port `8080` to the machine port.
- `docker-compose.yml` needs updating from version `2` to version `3` as the current version of `docker-compose` is now `3` and is recommended by Docker official site.


aptarmy:
Close because machine port conflicts with default redirect URL defined in Oauth2 server(facebook, google etc.)

PR #791, opened 2018-03-16
fix - returnTo with queryParams

### LINKED ISSUE #789: returnTo - does not work as expected for urls with query params
`req.session.returnTo = req.path;` will actually ignores the query params
can we change it to `req.session.returnTo = req.url;`? 

Ref: https://github.com/sahat/hackathon-starter/blob/master/app.js#L109

shreedharshetty:
If you want to redirect to some url with queryParams after sign up/login, app only considers path. as we set the `req.path` to `req.session.returnTo`.
This will set `req.url` so path with queryParams will be set to `req.session.returnTo`

fixes #789

### DIFF
--- a/app.js
+++ b/app.js
@@ -106,10 +106,10 @@ app.use((req, res, next) => {
       req.path !== '/signup' &&
       !req.path.match(/^\/auth/) &&
       !req.path.match(/\./)) {
-    req.session.returnTo = req.url;
+    req.session.returnTo = req.originalUrl;
   } else if (req.user &&
       req.path === '/account') {
-    req.session.returnTo = req.url;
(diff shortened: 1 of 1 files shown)


YasharF:
Is there a reason on why `req.path` is prefered over `req.originalUrl` or does `req.originalUrl` is a better fit?

Docs:

- http://expressjs.com/en/api.html#req.path
- http://expressjs.com/en/api.html#req.originalUrl

shreedharshetty:
👍 @YasharF as req.path is not intended to capture query params and since req.url is not native express property I think req.originalUrl makes sense

YasharF:
Thank you.

PR #796, opened 2018-03-21
Updates to README.md

YasharF:
- Fixed broken images in README.md
- Fixed mlab's link in README.md
- Replaced multiple references to port 3000 with port 8080 which is currently the default port defined in app.js
- Clarified that the Docker will be at the localhost port 8080 by default.
- Updated the copyright date in license to include 2018.


YasharF:
I also removed the Thinkful badge, since the service didn't seem to be available anymore.

PR #804, opened 2018-03-27
better .gitignore

nacimgoura:
Based on https://github.com/elsewhencode/project-guidelines/blob/master/.gitignore


YasharF:
What is each line, and why is it being added?
For example, I saw .ipr added, which is for InstallShield Professional Project File. Why would Hachathon starter end up with an ipr file?

.gitignore
```diff
@@ -25,8 +28,24 @@ bower_components
...
 # OS metadata
 .DS_Store
 Thumbs.db
+.DocumentRevisions-V100
+.fseventsd
+.Spotlight-V100
+.TemporaryItems
+.Trashes
+.VolumeIcon.icns
+.com.apple.timemachine.donotpresent
```
YasharF: please add an end of line on the last line

nacimgoura:
The jetbrains IDE use and generate ipr file. All other added rows are for files that can be added by the operating system.

PR #805, opened 2018-03-27
better .eslintrc

nacimgoura:
Adding env in eslint allows better code detection. For example `describe` is not underlined in red with this configuration because eslint detects that it is a unit test library.
Also, eslint always displayed `Definition for rule` chai-friendly/no-unused-expressions' was not found` because the plugin was not declared in.eslintrc.


YasharF:
Thank you

Issue #720, opened 2017-08-11
Refactor using ES6 promises & async/await


mayeaux:
Promises are a lot more readable and fun to work with. Any plans to refactor with ES6 promises and async/await, would you be open to that for the project?

niallobrien:
There's quite a difference between promises and async/await with regards to syntax and this projects is already utilising promises. Have you noticed somewhere where promises aren't being utilised? If so, it would make a great PR.
Regarding async/await - it would require a lot more changes and further consensus from the community as to whether most would want to use try/catch or use a general catch method in a wrapper and use that wrapper instead.

mrmayfield:
Thanks for the response. Maybe I'm crazy but everything looks to be done using callbacks to me. For example: https://github.com/sahat/hackathon-starter/blob/master/controllers/user.js#L36

I'm pretty sure Mongoose returns promises by default so they should be thenable. 

As for async/await, it's been shipped in native node for a while and makes code noticeably more readable & easy to interact with, especially for novices (who struggle with the callback concept, also there's no more callback spaghetti).

Am I really the first to have brought up using async/await? 😆

niallobrien:
Please submit a PR for any code improvements you can contribute.
Have you seen this previous PR? https://github.com/sahat/hackathon-starter/pull/638

Regarding async/await, no you're not. See https://github.com/sahat/hackathon-starter/issues/698

slidenerd:
You could always make another branch with async await :) and link the README.md to it, i didnt want to open a new issue, hence a comment here, the main.scss file is using an outdated layout it seems, would it be possible for you to add a gist on the same layout done with a flexbox/css grid instead Thanks

YasharF:
If anyone is interested to help with PRs, an intermediate task in the backlog is also:

- Remove bluebird package since Node.js has supported promises for a while, and Node.js 8 now supports utils.promisify() to turn callbacks into promises.

This could be done independently before a larger refactoring, since bluebird is only used in a few places.

nacimgoura:
@YasharF I can watch that this weekend. 😃

PR #807, opened 2018-03-27
use new es6 syntax

nacimgoura:
With recent versions of nodeJS, it is possible to use the template litteral and destructuring assignment. I also used eslint to correct possible syntax errors.
Link with #720

### DIFF
--- a/config/passport.js
+++ b/config/passport.js
@@ -8,8 +8,8 @@ const GitHubStrategy = require('passport-github').Strategy;
 const GoogleStrategy = require('passport-google-oauth').OAuth2Strategy;
 const LinkedInStrategy = require('passport-linkedin-oauth2').Strategy;
 const OpenIDStrategy = require('passport-openid').Strategy;
-const OAuthStrategy = require('passport-oauth').OAuthStrategy;
-const OAuth2Strategy = require('passport-oauth').OAuth2Strategy;
+const { OAuthStrategy } = require('passport-oauth');
+const { OAuth2Strategy } = require('passport-oauth');
 
 const User = require('../models/User');
 
@@ -405,15 +405,16 @@ passport.use(new InstagramStrategy({
 /**
  * Tumblr API OAuth.
  */
-passport.use('tumblr', new OAuthStrategy({
-  requestTokenURL: 'http://www.tumblr.com/oauth/request_token',
-  accessTokenURL: 'http://www.tumblr.com/oauth/access_token',
-  userAuthorizationURL: 'http://www.tumblr.com/oauth/authorize',
-  consumerKey: process.env.TUMBLR_KEY,
-  consumerSecret: process.env.TUMBLR_SECRET,
-  callbackURL: '/auth/tumblr/callback',
-  passReqToCallback: true
-},
+passport.use('tumblr', new OAuthStrategy(
+  {
+    requestTokenURL: 'http://www.tumblr.com/oauth/request_token',
+    accessTokenURL: 'http://www.tumblr.com/oauth/access_token',
+    userAuthorizationURL: 'http://www.tumblr.com/oauth/authorize',
+    consumerKey: process.env.TUMBLR_KEY,
+    consumerSecret: process.env.TUMBLR_SECRET,
+    callbackURL: '/auth/tumblr/callback',
+    passReqToCallback: true
+  },
   (req, token, tokenSecret, profile, done) => {
     User.findById(req.user._id, (err, user) => {
       if (err) { return done(err); }
@@ -428,14 +429,15 @@ passport.use('tumblr', new OAuthStrategy({
 /**
  * Foursquare API OAuth.
  */
-passport.use('foursquare', new OAuth2Strategy({
-  authorizationURL: 'https://foursquare.com/oauth2/authorize',
-  tokenURL: 'https://foursquare.com/oauth2/access_token',
-  clientID: process.env.FOURSQUARE_ID,
-  clientSecret: process.env.FOURSQUARE_SECRET,
-  callbackURL: process.env.FOURSQUARE_REDIRECT_URL,
-  passReqToCallback: true
-},
+passport.use('foursquare', new OAuth2Strategy(
+  {
+    authorizationURL: 'https://foursquare.com/oauth2/authorize',
+    tokenURL: 'https://foursquare.com/oauth2/access_token',
+    clientID: process.env.FOURSQUARE_ID,
+    clientSecret: process.env.FOURSQUARE_SECRET,
+    callbackURL: process.env.FOURSQUARE_REDIRECT_URL,
+    passReqToCallback: true
+  },
   (req, accessToken, refreshToken, profile, done) => {
     User.findById(req.user._id, (err, user) => {
       if (err) { return done(err); }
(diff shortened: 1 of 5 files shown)


controllers/contact.js
```diff
@@ -42,10 +39,9 @@ exports.postContact = (req, res) => {
     return res.redirect('/contact');
   }
 
-  if (!req.user) {
-    fromName = req.body.name;
-    fromEmail = req.body.email;
-  } else {
+  let fromName = req.body.name;
```
huttj: Is this endpoint re-used? I think it was more clear the other way you had it.

huttj:
Looks cleaner. Not a fan of the line breaks in the function arguments, but that's just a style thing.

nacimgoura:
on the contrary, I find it clearer now. Also, it is necessary to assign a value to a variable declared with let, that's why I directly assign the value to it. For the rest I simply respected the rules as they were defined in the project.

huttj:
Well, my thinking is that where the values come from (form or session) depends on whether or not the session is defined. With an if/else, it's clear that only one of the branches is executed, and that they are mutually exclusive. Of course, the outcome is the same, and it is kind of a matter of personal preference, so I think it's fine the way you have it.

You _can_ declare a variable with `let` without assigning a value, so you could keep the same structure as before, if you wanted to.

YasharF:
I agree with @huttj that the if/else is more clear.  When writing the original code for this part I actually thought about both ways of having an if/else or initializing and then conditionally assigning and the final deciding factor was to opt for easy reading.
I'm not sure if the engine does any optimization, but in low level languages, you are spending more CPU cycles if the code is assigning a value followed by an if condition to conditionally assign a new value to the same variable compared to the case that there is an if/else to do value assignment only once.

controllers/api.js
```diff
@@ -287,15 +287,18 @@ exports.postTwitter = (req, res, next) => {
...
   const steamId = req.user.steam;
   const params = { l: 'english', steamid: steamId, key: process.env.STEAM_KEY };
-  const playerAchievements = () => {
-    //get the list of the recently played games, and pick the most recent one and get its achievements
-    return request.getAsync({ url: 'http://api.steampowered.com/IPlayerService/GetRecentlyPlayedGames/v0001/', qs: params, json: true })
+  const playerAchievements = () =>
+    /**
+     * get the list of the recently played games,
+     * and pick the most recent one and get its achievements
+     */
```
YasharF: Was this comment getting broken up caused by eslint?

config/passport.js
```diff
@@ -428,14 +429,15 @@ passport.use('tumblr', new OAuthStrategy({
...
-passport.use('foursquare', new OAuth2Strategy({
-  authorizationURL: 'https://foursquare.com/oauth2/authorize',
-  tokenURL: 'https://foursquare.com/oauth2/access_token',
-  clientID: process.env.FOURSQUARE_ID,
-  clientSecret: process.env.FOURSQUARE_SECRET,
-  callbackURL: process.env.FOURSQUARE_REDIRECT_URL,
-  passReqToCallback: true
-},
+passport.use('foursquare', new OAuth2Strategy(
+  {
```
YasharF: Same as what @huttj mentioned - not sure about the line break.  Most of the rest of the project doesn't have such a line break before the '{'.  If eslint rules are causing this, we may want to look at why and if there is something in the eslint rules that need to be fixed.

YasharF: It doesn't look like that we have it configured and it is probably going to its default.  Perhaps we need to add `"function-paren-newline": ["error", "never"],` to the eslint configs.

doc: https://eslint.org/docs/rules/function-paren-newline

controllers/api.js
YasharF: It looks like the lit rules marked this because the comment line was more than 100 char long.  It was 103 char long.  I would rather just change one of the `and`sto `&` and make it conform to the 100 char limit than split it into multiple lines in order to keep the code compact.

YasharF:
Now that eslint is working, it looks like there are five other lines that could get some ES6 love; do you think if you can add those to this PR as well?

```
/mnt/c/temp/nac/hackathon-starter/controllers/api.js
  364:9   error  Use object destructuring  prefer-destructuring
  365:9   error  Use object destructuring  prefer-destructuring
  519:11  error  Use object destructuring  prefer-destructuring
  535:9   error  Use object destructuring  prefer-destructuring

/mnt/c/temp/nac/hackathon-starter/controllers/user.js
  202:9  error  Use object destructuring  prefer-destructuring
```

nacimgoura:
It's done :wink:

YasharF:
Awesome.  I fixed up the eslint configs.  Just as soon as the following two are in I'll have it merge:


1. The line breaks in the function calls which the eslint configs were not configured right for are now being cough with the latest eslint config on master:
```
hackathon-starter/config/passport.js
  408:41  error  Unexpected newline after '('   function-paren-newline
  427:1   error  Unexpected newline before ')'  function-paren-newline
  432:46  error  Unexpected newline after '('   function-paren-newline
  450:1   error  Unexpected newline before ')'  function-paren-newline
  519:45  error  Unexpected newline after '('   function-paren-newline
  537:1   error  Unexpected newline before ')'  function-paren-newline
```

2. Reverting the assignments for fromName and fromEmail to the prior if/else format. https://github.com/sahat/hackathon-starter/blob/master/controllers/contact.js#L45  like the following.  Although, we can potentially change fromName an fromEmail to name and email and utilize ES6 formatting (make sure to test it if renaming to name and email).
```
  if (!req.user) {
    fromName = req.body.name;
    fromEmail = req.body.email;
  } else {
    fromName = req.user.profile.name || '';
    fromEmail = req.user.email;
  }
```


Thank you again, getting all of the ES6 syntax really needed some love and I hope we can merge this tomorrow or so.

nacimgoura:
I'm revert for if/else format, even though I found it clearer and more concise by directly affecting the variable. I would have preferred to do it like that, with a completely es6 structure.

```
exports.postContact = (req, res) => {
  const { body, user } = req;
  let { name: fromName, email: fromEmail } = body;
  if (!user) {
    req.assert('name', 'Name cannot be blank').notEmpty();
    req.assert('email', 'Email is not valid').isEmail();
  } else {
    ({ profile: { name: fromName = '', email: fromEmail } } = req.user);
  }
  req.assert('message', 'Message cannot be blank').notEmpty();

  const errors = req.validationErrors();

  if (errors) {
    req.flash('errors', errors);
    return res.redirect('/contact');
  }

  const mailOptions = {
    to: 'your@email.com',
    from: `${fromName} <${fromEmail}>`,
    subject: 'Contact Form | Hackathon Starter',
    text: body.message
  };

  transporter.sendMail(mailOptions, (err) => {
    if (err) {
      req.flash('errors', { msg: err.message });
      return res.redirect('/contact');
    }
    req.flash('success', { msg: 'Email has been sent successfully!' });
    res.redirect('/contact');
  });
};
```

huttj:
>    ({ profile: { name: fromName = '', email: fromEmail } } = req.user);

Oh, is that how you do that? Very clever! I've wanted to be able to destructure into existing variables for a long time. Awesome.

@nacimgoura: I like that latest approach, because the validation and data wrangling happen at the top, and then the actual "business logic" at the bottom. Nice and clean.

YasharF:
It is definitely compact and elegant code.  What made me hesitate though was that I rather see input sanitization done before even variable initialization using unsanitized inputs.

PR #390, opened 2016-03-01
Add Pinterest OAuth and API examples.

theGrue:
Added Pinterest OAuth configuration and API examples. Examples show how to get a user's boards and add pins to those boards. Closes #381.

PR #812, opened 2018-03-30
add https support

### LINKED ISSUE #504: Pinterest API example have invalid redirect URL
.env.example, last line:
PINTEREST_REDIRECT_URL=https://localhost:3000/auth/pinterest/callback

should be http instead of https

Not sure if i should make PR for this one-line fix.

nacimgoura:
For some api, it is necessary that the site is in https, for example with pinterest. The pinterest api is not usable without https. Fix #504

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,7 @@ Thumbs.db
 .Trashes
 .VolumeIcon.icns
 .com.apple.timemachine.donotpresent
+
+# Certificate
+*.key
+*.crt
--- a/README.md
+++ b/README.md
@@ -132,6 +132,9 @@ npm install
 
 # Then simply start your app
 node app.js
+
+# Fort start you app on https
+node app.js HTTPS
 ```
 
 **Note:** I highly recommend installing [Nodemon](https://github.com/remy/nodemon).
--- a/app.js
+++ b/app.js
@@ -1,7 +1,10 @@
 /**
  * Module dependencies.
  */
+const fs = require('fs');
 const express = require('express');
+const http = require('http');
+const https = require('https');
 const compression = require('compression');
 const session = require('express-session');
 const bodyParser = require('body-parser');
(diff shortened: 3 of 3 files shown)


YasharF:
Three questions that come to my mind before merging this:
1. How does this add complexity for a hackathon team during a hackathon weekend?  Specially considering that many users may not have a whole lot of experience with NodeJS, Express, etc.
2. If this is going to help them instead of adding complexity or potential things that would block them, then we need some kind of instructions or URL to instructions on how to create at least a self-signed certificate since otherwise the app will crash when trying to run node app.js HTTPS.  Which also means, there needs to be some error handling in the part of the code that reads the certs, so a better error message is displayed to help them add the cert instead of `fs.js:646 return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);` error that appears now.  Would a self-signed certificate work, or is that something that will have its own set of problems?
3. Does this create any issues with the existing hosting services that we have listed?  For example, Heroku natively bind the app to HTTPS without this change using Heroku's own platform https certs.  Will this cause issues with Heroku or other platforms?

nacimgoura:
I haven't thought about any of this. I just saw that it was impossible to use the pinterest api locally because of that.

1) Yes it adds complexity because you need to have knowledge of ssl certificates.

2) To generate a self-signed certificate, I read this tutorial: https://devcenter.heroku.com/articles/ssl-certificate-self

3) I don't know but I don't think that's a problem.

Let's leave that aside then.

YasharF:
For local/development testing in the instagram case, can https://ngrok.com be used as described in https://github.com/sahat/hackathon-starter/pull/390#issuecomment-190911282 without this patch?

nacimgoura:
I just tested with ngrok, the pinterest api works correctly.

YasharF:
We should probably add something to the Readme about ngrok.  It would be hard to dig through closed issues to figure it out.

Issue #698, opened 2017-05-12
Use async/await instead of promises (Node 7.6+)


plegner:
ES6 [async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) is supported by Node 7.6+ with [V8 5.5+](https://v8project.blogspot.co.uk/2016/10/v8-release-55.html). This allows many of the promise-based functions to become much simpler and more readable.

For example, the [postReset](https://github.com/sahat/hackathon-starter/blob/master/controllers/user.js#L242) function would simplify to:

```javascript
exports.postReset = async function(req, res, next) {
  // ...

  try {
    const user = await User.findOne({passwordResetToken: req.params.token})
        .where('passwordResetExpires').gt(Date.now());

    if (!user) { /* ... */ }

    user.password = req.body.password;
    user.passwordResetToken = undefined;
    user.passwordResetExpires = undefined;

    await user.save();
    await req.logIn(user);

    const transporter = // ...
    const mailOptions = // ...
    await transporter.sendMail(mailOptions);

    req.flash('success', {msg: 'Success! Your password has been changed.'});
    if (!res.finished) res.redirect('/');

  } catch (e) {
    next(e);
  }
}
```

niallobrien:
Maybe we should get a PR together for this?

niallobrien:
We'd most likely need a wrapper function which would do the try/catch so we wouldn't have to litter the entire codebase with such blocks.

Kessir:
Hi Everyone, any update  on this?
I totally agree with the idea of using a wrapper.

YasharF:
My 2 cents is to hold off on this since if I understood the original posting correctly, this will mean that the project will no longer be compatible with NodeJS versions prior to 7.6.

The current LTS version of NodeJS is Node 6; Node 7 will never be an LTS version, and Node 8 is slated to become LTS later this month.  Assuming we wait for NodeJS 8's LTS status, my other 2 cents would be to wait for a few months for the underlying dependencies and other parts of the ecosystem to have a chance to move from Node 6 to Node 8 at least, and then draw the line in the sand and migrate the project with a major version increment.

Kessir:
@YasharF That's a fair point.

YasharF:
NodeJS 8.x branch is now the latest LTS.  Has anyone done testing of HackatonStarter on NodeJS 8?

niallobrien:
[...]

PR #814, opened 2018-04-01
add async/await, es6 structure and multiple improvements

nacimgoura:
I've made many changes:
- Remove bluebird dependencies
- Replace request by axios
- Reformat code with es6 syntax
- Use async/await
- Clearer code
- Multiple improvements
- Update README

There is still code that can be refactorized but I did most of it.

Due to many changes I could not test everything. I did `npm test`, and test the overall functioning of the application and everything seems to work. 

Relative issues : #720, #698

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
     "multer": "^1.3.0",
     "node-foursquare": "^0.3.2",
     "node-linkedin": "^0.5.6",
-    "node-sass": "^4.7.2",
     "node-sass-middleware": "^0.11.0",
     "nodemailer": "^4.6.3",
     "passport": "^0.4.0",
--- a/package.json
+++ b/package.json
@@ -16,8 +16,8 @@
   },
   "dependencies": {
     "@octokit/rest": "^15.2.5",
+    "axios": "^0.18.0",
     "bcrypt-nodejs": "^0.0.3",
-    "bluebird": "^3.5.1",
     "body-parser": "^1.18.2",
     "chalk": "^2.3.2",
     "cheerio": "^1.0.0-rc.2",
@@ -56,7 +56,6 @@
     "passport-twitter": "^1.0.4",
     "paypal-rest-sdk": "^1.8.1",
     "pug": "^2.0.3",
-    "request": "^2.85.0",
     "stripe": "^5.5.0",
     "tumblr.js": "^1.1.1",
     "twilio": "^3.14.0",
--- a/README.md
+++ b/README.md
@@ -360,6 +360,7 @@ List of Packages
 | Package                         | Description                                                           |
 | ------------------------------- | --------------------------------------------------------------------- |
 | @octokit/rest                   | GitHub API library.                                                   |
+| axios                           | Promise based HTTP client for the browser and node.js                 |
 | bcrypt-nodejs                   | Library for hashing and salting user passwords.                       |
 | body-parser                     | Express 4 middleware.                                                 |
 | chai                            | BDD/TDD assertion library.                                            |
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -41,7 +41,6 @@
     "multer": "^1.3.0",
     "node-foursquare": "^0.3.2",
     "node-linkedin": "^0.5.6",
-    "node-sass": "^4.7.2",
```
YasharF: Why is nore-sass being removed?

nacimgoura:
I thought I had detailed it. Because it's already declared in peerDependencies.

nacimgoura:
Thank you for your feedback, but I don't understand some things :
- The only modification for controllers/home.js is the es6 structure with implicit return.
- User model doesn't changes (there is just a line break for clarity but no impact on operation).
- I remove node-sass because it's declared twice in package.json.
- 
```
Console.log messaging when the server starts, out of scope. There are issues with that line as it currently is, but there is more work that needs to be done to fix that line.
```
I don't understand this sentence, can you give me more details?


I am not submitting untested changes. I run an `npm test` and test myself before publishing. But there are some APIs (like instagram) that require external validation and so I can't test them.

From now on I will make small pull requests.

YasharF:
> The only modification for controllers/home.js is the es6 structure with implicit return.
> User model doesn't changes (there is just a line break for clarity but no impact on operation).
> console.log messaging when the server starts

From the standpoint of a reviewer or a post-hackathon user (doing rebase or picking diffs), why was the original line _broken_ in the first place requiring a change?  This is specially important from a post-hackathon user considering that there might be merge conflicts that they need to figure out.  Is it addressing a bug?  Is it something that they need to look at the rest of their code and replicate throughout?  Is it just code-style change?  If it is code-style, why was it not important enough to be flagged by eslint?  Can that change by itself justify its need in a standalone PR?  For example, is there currently no implicit return in the specific place, and what is the problem that it is causing?
When having smaller commits and PRs that are focused on specific problems instead of a bundled case, these can be clarified one by one.  Take a look at the commit description for https://github.com/sahat/hackathon-starter/commit/e9f0e8e673cbf9b0f8dd0aef808da14403ab3b4b ; even though there are only 7 lines of code that have changed, there is a much longer commit description (158 words) that has been added to justify why the change is needed and without it the code base has bugs.

> I remove node-sass because it's declared twice in package.json

What problem was this causing?  Was there a warning in npm?  Since this involves the build process, it requires verification that it doesn't break builds on supported stacks: Heroku, OpenShift, Azure, Bluemix, GCP

> I am not submitting untested changes. I run an npm test and test myself before publishing. But there are some APIs (like instagram) that require external validation and so I can't test them.

Currently there is no coverage for most issues with the current test cases, and `npm test` is not sufficient for testing.  Most changes do need to be manually tested.  In case of the APIs, they need to be tested using real keys and real accounts for various cases that the API example may run into.  In other words, you do need to run the server and make sure that no functionality.  Another benefit of having smaller focused PRs is that, since you need to manually test your changes, you can focus on a smaller number of manual checks that cover the scope of your changes before submitting the PR.

nacimgoura:
Okay, I understand. Do not merge or close this pull request, I leave it there to serve as reference. After this, we can close it.

YasharF:
Is there anything left on this to do or can we close this PR?

nacimgoura:
Yes, there's still :
- use of promises instead of callbacks for mongo requests
- possibly the use of axios instead of request

YasharF:
Lets reduce the scope to just 

> use of promises instead of callbacks for mongo requests

so we can close this, and we'll revisit RequestJS vs Axios at a later point.

PR #815, opened 2018-04-01
add mongo-express

nacimgoura:
add [mongo-express](https://github.com/mongo-express/mongo-express) for visualize mongoDB database content in browser.

mongo-express


YasharF:
A stand-alone package such as MongoDB Compass can be used to access the DB for this function, without having to add it to the docker image.  I rather not bloat the project as much as possible.

Issue #810, opened 2018-03-29
bug api steam


nacimgoura:
Hello,
When trying to use the api steam, I encounter this error.
```
Connect
500 TypeError: Cannot read property 'playerstats' of undefined
   at Promise.all.then (/Users/nacim/Documents/project/hackathon-starter/controllers/api.js:338:49)
   at <anonymous>
   at process._tickCallback (internal/process/next_tick.js:188:7)
```

The problem comes from this part of the code.
```
Promise.all([
    playerAchievements(),
    playerSummaries(),
    ownedGames()
  ])
    .then(([playerAchievements, playerSummaries, ownedGames]) => {
      res.render('api/steam', {
        title: 'Steam Web API',
        ownedGames: ownedGames.response.games,
        playerAchievemments: playerAchievements.playerstats,
        playerSummary: playerSummaries.response.players[0]
      });
    })
    .catch(next);
```

YasharF:
Good catch.  It looks like that there is a missing error handling.  If the player doesn't have any recently played game or achievements, the object is undef but pug still makes an attempt to render it causing it to crash.

YasharF:
https://github.com/sahat/hackathon-starter/pull/811

YasharF:
Fixed with https://github.com/sahat/hackathon-starter/commit/525e089a053bfcfd4d5cdfce8008609f4724820f

PR #818, opened 2018-04-01
Remove bluebird

nacimgoura:
bluebird is useless with nodejs 8 because promises are used natively.

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -1,8 +1,10 @@
-const bluebird = require('bluebird');
-const crypto = bluebird.promisifyAll(require('crypto'));
+const { promisify } = require('util');
+const crypto = require('crypto');
 const nodemailer = require('nodemailer');
 const passport = require('passport');
 const User = require('../models/User');
+
+const randomBytesAsync = promisify(crypto.randomBytes);
 
 /**
  * GET /login
@@ -329,8 +331,7 @@ exports.postForgot = (req, res, next) => {
     return res.redirect('/forgot');
   }
 
-  const createRandomToken = crypto
-    .randomBytesAsync(16)
+  const createRandomToken = randomBytesAsync(16)
     .then(buf => buf.toString('hex'));
 
   const setRandomToken = token =>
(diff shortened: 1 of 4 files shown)


controllers/user.js
```diff
@@ -1,9 +1,11 @@
-const bluebird = require('bluebird');
-const crypto = bluebird.promisifyAll(require('crypto'));
+const { promisify } = require('util');
+const crypto = require('crypto');
 const nodemailer = require('nodemailer');
 const passport = require('passport');
 const User = require('../models/User');
 
+const randomBytesAsync = promisify(crypto.randomBytes);
```
YasharF: Can we have the same compact one line definition as before?  Why or why not?

nacimgoura: This is not possible according to the [documentation](https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original).

YasharF:
If you haven't already. can you do the following functional tests manually in the browser and report here?

- Does the password reset process still work?  Asking, since I think crypto related code for that has been touched.  Can you verify in the DB during the process that there is in fact a randomized token present?
- Can you manually verify in the browser that Instagram, FourSquare, and Steam API examples still work, while using working API keys from the three providers, and using real accounts.  You may need more than one account if there are edge cases (aka like the Steam achievement issue in the other PR).

nacimgoura:
I ran the tests again.
the password reset process generates a value like this `5306492510cc17dc16e7040c54488f90`.

But for steam I had created my account specifically to test the project, so I have no achievement, which allowed me to find bug #810. Since I still don't have any achievements and I don't have an account with achievements, I can't test this particular case.

YasharF:
I retested for Steam, Foursquare, and password reset.  No issues.

I did not test for Instagram, due to issues that I ran into with getting permissions through Instagram developer portal, and going to skip testing it since Hackathon Starter readme doesn't even cover it,  Instagram's current APIs are on the deprecation path, and that the npm package is unmaintained.

Thank you!

YasharF:
Note: in-part to replace https://github.com/sahat/hackathon-starter/pull/814

PR #820, opened 2018-04-01
Fix bug display image in view instagram

nacimgoura:
For the instagram page the images are not displayed. HTML display '#{user.profile_picture}' instead of value.


YasharF:
Thank you!

PR #828, opened 2018-04-02
remove duplicate node-sass

nacimgoura:
node-sass is already declared in dependencies but it is also declared in PeerDependencies. It should be removed from PeerDependencies.
Also the version in peerDependencies has not been updated.
The library is even ignored when generating the package-lock.json file because it takes into account the last version declared in the dependencies.
I tested in local, travis build and heroku. It's working fine. 
In-part from #814

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,9 +62,6 @@
     "twit": "^2.2.9",
     "validator": "^9.4.1"
   },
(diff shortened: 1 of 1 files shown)


YasharF:
Thank you!

PR #831, opened 2018-04-02
refactor github api with es6 syntax and async/await

nacimgoura:
github library ([@octokit/rest](https://github.com/octokit/rest.js)) natively supports promise.

### DIFF
--- a/controllers/api.js
+++ b/controllers/api.js
@@ -124,12 +124,12 @@ exports.getGithub = async (req, res, next) => {
   const github = new GitHub();
   try {
     const { data: repo } = await github.repos.get({ owner: 'sahat', repo: 'hackathon-starter' });
-    return res.render('api/github', {
+    res.render('api/github', {
       title: 'GitHub API',
       repo
     });
   } catch (error) {
-    return next(error);
+    next(error);
   }
 };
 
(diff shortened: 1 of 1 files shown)


YasharF:
I haven't merged this yet, since it rang some bells in my mind regarding the return keyword in front of res.render.  I couldn't find anything that it is required in ES6 and without it I think everything should work, right?  Asking a few people around, it looks like it is more or less of a code style item.  Perhaps https://eslint.org/docs/rules/consistent-return .  So the bigger question is that should we set a rule to preserve the prior no-return required state, or should we change the project's eslint rule to require such return statements.

YasharF:
> when there is the ’async’ keyword at the beginning of a function, it's recommended to have ’return’ at the end of the function.

I haven't been able to find any official documentation to confirm that we need to do this yet.  I think it is adding extra fluff to the coding as is.  I am still researching this.  I also wasn't able to find a ESLint rule that specifically relates to this case.

YasharF:
I have not found any official documentation that the `return` for the `res.render` is required.  In favor of simplicity (less wordiness), lets remove the `return` that is for the `res.render` statement for now.  We may revisit this at a later point.

YasharF:
Thank you :)

YasharF:

PR #836, opened 2018-04-07
update README.md for the migration from jade to pug

nacimgoura:
There were breaks between the transition from jade to pug (https://pugjs.org/api/migration-v2.html). Some of the links in the readme point to an old documentation or old converter. I've updated all this.


YasharF:
Thank you

PR #837, opened 2018-04-08
update README.md with import/export supported by node

nacimgoura:
It describes how to import and export with the es modules, but they are not supported by current versions of nodeJS. It's [experimental](https://nodejs.org/api/esm.html#esm_ecmascript_modules). You can use them with babel (or another transpilator), but it is not included in the project and will add a layer of complexity. I updated the readme with the currently supported native syntax.


sahat:
That's a good idea. 

For reference: https://nodejs.org/api/esm.html

PR #839, opened 2018-04-09
refactor twitter api with es6 syntax and async/await

nacimgoura:
[Twit](https://github.com/ttezel/twit) library natively supports promise. It is better to use the new syntax es6. This eliminates the need for callback.


YasharF:
Thank you :)

YasharF:

PR #843, opened 2018-04-28
reformat partially getSteam with es6 syntax

nacimgoura:
Partial reformatting of the getSteam function with ES6 syntax. Fixed a bug that could happen if you had no game. Also, I have the impression that the api steam has recently changed but I haven't done any more research on it.


YasharF:

PR #845, opened 2018-04-28
remove mongoose global.Promise

nacimgoura:
`mongoose.Promise = global.Promise` is useless with mongoose > 5 because it supports natively Promise.


YasharF:
Ref: http://thecodebarbarian.com/introducing-mongoose-5.html

PR #846, opened 2018-04-28
call only promisify on util

nacimgoura:
Calling only the function that is used allows more clarity and lightness.
Replace `const util = require('util');`by `const { promisify } = require('util');`


YasharF:
Thank you!

Issue #766, opened 2017-12-09
Bootstrap 4 and Font Awesome 5


sahat:
Creating an issue to track these tasks.

- [x] Bootstrap 4 beta
- [x] Font Awesome 5

slidenerd:
![image 1](https://user-images.githubusercontent.com/5139030/33883453-d5c8ca12-df61-11e7-82aa-b1d8cfe94a5a.png)
Looks like the UI classes in bootstrap 4 beta 2 have changed significantly :) and bootstrap social doesnt work either

imcodingideas:
@slidenerd because scss is awesome.

slidenerd:
@imcodingideas  the reason this feels weird is because the repo uses sass loader for scss which is pretty nice but has nothing on js, i see a bunch of js min files scattered and thats it, while i am not the greatest fan of webpack, i was wondering why the repo doesnt use cdnjs urls with a local fallback to load all js files plus babel loaders and presets, all we want in the end is cross browser es6 code that loads super fast from a cdn url with a local fallback and loads only 1 file

niallobrien:
I implemented Laravel Mix for the front-end asset pipeline in my fork which does scss and js compilation. It also has a production option to serve optimised assets.

YasharF:
PR for Font Awesome 5 upgrade: https://github.com/sahat/hackathon-starter/pull/799

YasharF:
For Bootstrap 4 upgrade, I would suggest replacing the /public/css/lib/bootstrap with the npm distributed Bootstrap which includes the scss files.  In that way, we'll be dealing with less effort when there are minor version updates to Bootstrap 4, since an `npm update` would pick up the change.

For Bootstrap Social, since the current maintainer has not been able to update it for Bootstrap 4, another community member has created a fork with the updates `@ladjs/bootstrap-social`.  The following would be my suggested diffs for package.json and main.scss changes to pick up these distributions:


```
diff --git a/package.json b/package.json
index b0ee039..879cb60 100644
--- a/package.json
+++ b/package.json
@@ -9,16 +9,18 @@
   "author": "Sahat Yalkabov",
   "license": "MIT",
   "scripts": {
-    "postinstall": "npm rebuild node-sass",
+    "postinstall": "cp node_modules/bootstrap/dist/js/bootstrap.min.* public/js/lib && npm rebuild node-sass",
     "start": "node app.js",
     "test": "nyc mocha --timeout=3000 --exit",
[...]

PR #847, opened 2018-04-29
Bootstrap 4 Upgrade

YasharF:
This merge includes:
- Upgrade of all of the views to be compatible with bootstrap 4 
- Upgrade of Bootstrap and jQuery.  The libraries are now being installed using their npm package to reduce manual efforts for minor version upgrades.
- Changes to default theme and main.scss to allow the site to load with only minor visual differences between bootstrap 3 and bootstrap 4.
- Upgrade of flatly theme to its bootstrap 4 version.


YasharF:
Known issues:

-  Modern and GSDK themes have not been updated.
-  The issues listed at https://github.com/sahat/hackathon-starter/issues/766#issuecomment-385077977

Most likely I would need some help to fix these remaining issues.  If there are no comments in the next ten days or so, I'll merge this as is, and we'll fix the remaining issues in future minor releases.

PR #858, opened 2018-05-14
chore: clone last 5 commits

DanielRuf:
By default Travis clones the last 50 commits. 5 are sufficient and often faster.

See https://docs.travis-ci.com/user/customizing-the-build#Git-Clone-Depth

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ node_js:
   - "8"
   - "9"
   - "10"
+git:
+  checkout: 5
 services: mongodb  
 cache:  
   directories:


YasharF:
Is having the default of 50 creating any problems currently?  If so, why is the travis project not updating the default from their end?

DanielRuf:
> Is having the default of 50 creating any problems currently? If so, why is the travis project not updating the default from their end?

Not sure but it is the default. Many projects use a lower setting and both are shallow clones. No problems with 50 but longer checkouts in Travis builds.

YasharF:
Considering that the longer checkout doesn't impact the application performance, going to stay with the default.  If the default in general is a problem then the feedback should be provided to the Travis team.  If a specific application is having issues with the default then the change is mostly specific to that application.

Thanks for the PR though.

DanielRuf:
Well, faster CI builds = better DX but ok =)

YasharF:
Thanks.  If it is an impactful DX improvement, the feedback should be taken to the Travis folks to change the default.

PR #863, opened 2018-05-27
destructuring statuses from an extra layer deep to fix bug in view

garretthogan:
Do I need to add a test for this? If so, where at? There are currently only tests for the app and the models.


YasharF:
Thank you.

PR #873, opened 2018-06-25
Update README.md

Imadra:
Added useful express-generator module


YasharF:
express-generator is not a package dependency in hackathon-starter.

Issue #872, opened 2018-06-24
Home link in title bar is always active


dkimot:
As the title says, the Home link in the title bar is always active. If you delete the default active class from the `header.pug` file it solves this.

I'm happy to submit a PR, unless this is on purpose.

YasharF:
Is this what you are referring to?  If so, yes it looks like a bug that probably got introduced with the recent bootstrap 4 upgrade.  A PR would be awesome.  Thanks

![image](https://user-images.githubusercontent.com/6448697/41820825-8eecc9b8-778c-11e8-9df3-d5bb7da8b9f0.png)

dkimot:
Created a pull request.

PR #874, opened 2018-06-25
Fixed active class bug

dkimot:
Fixed a bug on the Home link's active class introduced by Bootstrap 4.


YasharF:
https://github.com/sahat/hackathon-starter/issues/872

Issue #877, opened 2018-06-28
postinstall script not working on windows because of cp


marcospgp:
I get the following error:

```
'cp' is not recognized as an internal or external command,
operable program or batch file.

**Comment by marcospgp** [comment]
This workaround fixed it for me:

https://github.com/ReactiveX/rxjs/issues/1601#issuecomment-214359513

**Comment by YasharF** [comment]
Can you submit a PR for the Readme.md to provide the issue with native windows nodejs installations?  Or would you be able to find out if there is an alternate way of getting those library assets to the right place?

These days I primarily use WSL Ubuntu on Windows 10 for all of my NodeJS development, but I suspect many people don't have my setup.

**Comment by marcospgp** [comment]
I use PowerShell which has cp out of the box, but npm seems to run on its
own cmd shell which doesn't.

I could add this to the readme, but I doubt most people would find it
there. Better to fix it silently and require no action by end users.

**Comment by YasharF** [comment]
It looks like there might be a better way than the hacky copy script if I am understanding one of the postings at https://stackoverflow.com/questions/26773767/purpose-of-installing-twitter-bootstrap-through-npm correctly.

```
var express = require('express');
var app = express();

// prepare server
app.use('/api', api); // redirect API calls
app.use('/', express.static(__dirname + '/www')); // redirect root
app.use('/js', express.static(__dirname + '/node_modules/bootstrap/dist/js')); // redirect bootstrap JS
app.use('/js', express.static(__dirname + '/node_modules/jquery/dist')); // redirect JS jQuery
app.use('/css', express.static(__dirname + '/node_modules/bootstrap/dist/css')); // redirect CSS bootstrap

_Then, finally, at the .html:_

<link rel="stylesheet" href="/css/bootstrap.min.css">
<script src="/js/jquery.min.js"></script>
<script src="/js/bootstrap.min.js"></script>
```

YasharF:
Fixed with https://github.com/sahat/hackathon-starter/commit/97538db2f8a3e0b429248d2b0163916d2e195667

PR #878, opened 2018-06-30
Using cpy-cli's cpy command instead of cp in postinstall

AlAskalany:
As mentioned in #877 the cp command causes an error when running npm install on Windows. To go around this I did the following:

- Installed the [cpy-cli](https://www.npmjs.com/package/cpy-cli) npm package as a DevDependency
- Replaced all cp commands in postinstall with cpy


YasharF:
Thanks for the PR.  Opted to use express static routes instead, to avoid increasing the number of dependencies.  npm audit is currently stating that there are 3047 packages in the dependency tree, which would be nice to trim if we can.

PR #883, opened 2018-07-26
Use Popper UMD

binarymax:
Popper was incorrectly referencing the version in root ```dist/```, and throwing an error when used in ```layout.pug```.  This now correctly references the UMD version in ```dist/umd/```


YasharF:
Thank you

PR #889, opened 2018-08-04
Update main.scss w/correct fontawesome filenames

gesa:
The 5.1.0 to 5.1.1 Font-Awesome update changed the names of those files to drop the `fa-` prefix. More information at their [upgrading](https://github.com/FortAwesome/Font-Awesome/blob/master/UPGRADING.md) guide.


YasharF:
Thank you!  I totally missed that from the 5.1.1 changelog:

> **Removed "fa-" prefix from Less and Sass style bundles filenames**

Issue #853, opened 2018-05-09
Bootstrap 4 upgrade issues


YasharF:
The following are known issues from the Bootstrap 4 upgrade, to be revisited:

- [x]  There is a 15px indentation in front of the nav-brand.  In the prior version, the responsive code was adding a -15px margin in some resolutions which was compensating for this.
![image](https://user-images.githubusercontent.com/6448697/39382248-bceced0c-4a19-11e8-8fa1-ce7bb595196a.png)

- [x]  navbar-header is deprecated.  The changes in the git commit causes the horizontal line to disappear when the nav items drop down with the navbar-toggler class.
![image](https://user-images.githubusercontent.com/6448697/39382344-013b5462-4a1a-11e8-9b33-c94ee6232892.png)

- [x]  There are some wrapping issues in the nav bar with some screen widths.
![image](https://user-images.githubusercontent.com/6448697/39382594-dd4657e0-4a1a-11e8-9721-09fe76753d23.png)

- [x] Fix navbar's theme issues (https://github.com/sahat/hackathon-starter/issues/853#issuecomment-475932533 and https://github.com/sahat/hackathon-starter/issues/853#issuecomment-475934766)
- [x] Modern theme upgrade or removal
- [x] GSDK upgrade or removal
- [x] Visual QA of flatly theme

estevanmaito:
Hi!
I would like to help you address these issues, but I don't know from where to start building the project. Would you mind pointing me the direction, as this is my first issue?
Thank you

estevanmaito:
@YasharF, thanks! I've already fixed points 1 and 3. 

Regarding point 2, what is the expected behavior? The collapsed bar is functioning as the examples from Bootstrap's site (take this as an example https://getbootstrap.com/docs/4.1/examples/album/), where the navbar expands and the bottom shadow follows.

And about the themes (points 4, 5 and 6), is there any discussion going on? I just changed the theme imports and there are some inconsistencies among them, like the navbar background being completely transparent and mixing with the text behind it. I don't know if I'm missing some import or what

```scss
/* imports that i've changed in order to change themes */
@import "themes/default/variables";
@import "themes/default/default";
```

YasharF:
I think the best way to do this would be to have two instances of the project running concurrently on two different ports. One using the Bootstrap 3 libraries as prior to the upgrade merge, and one as the current version which is using the Bootstrap 4 libraries.  You can create a clone of the repo with the state before BS4 changes by:
```
[...]

PR #890, opened 2018-08-04
Update broken file references

gesa:
`bootstrap-social` now lives in the node_modules directory, and is imported in main.scss. Meanwhile, this file seems to have lost importing its own variables, causing compile to fail.

### DIFF
--- a/public/css/themes/modern/_modern.scss
+++ b/public/css/themes/modern/_modern.scss
@@ -1,4 +1,4 @@
-@import "../../lib/bootstrap-social";
+@import "varibles";
 
 // Footer
 // -------------------------


YasharF:
The Modern Theme hasn't been upgraded ( https://github.com/sahat/hackathon-starter/issues/853 ).  If you are planning to use the modern theme, you may find other issues in there that may need to get fixed.  I would be happy to accept all related PRs to get that fixed up.  If you want, you can bundle the other Modern Theme fixes in this PR as well.

public/css/themes/modern/_modern.scss
```diff
@@ -1,4 +1,4 @@
-@import "../../lib/bootstrap-social";
```
YasharF: The reference here actually needs to point ot bootstrap social, and needs to be changed to `@import "node_modules/@ladjs/bootstrap-social/bootstrap-social.scss";`  that would be consistent with the way the other themes are.

For switching to different themes from default, you need to change two lines in main.scss:

- The variables: https://github.com/sahat/hackathon-starter/blob/master/public/css/main.scss#L1
- The scsss: https://github.com/sahat/hackathon-starter/blob/master/public/css/main.scss#L10

YasharF:
Closing this PR since it has been addressed by https://github.com/sahat/hackathon-starter/pull/894

PR #894, opened 2018-08-05
update the modern theme to be consistent with bootstrap

### LINKED ISSUE #893: Fail to load theme if change to Modern
If you change the theme to modern and force update css, there is a library error. 

How to reproduce:
1. force update of CSS file
```
app.use(sass({
  src: path.join(__dirname, 'public'),
  dest: path.join(__dirname, 'public'),
  force: true,
  debug: true
}));
```
2. change theme to modern
```
@import "themes/modern/variables";
...
@import "themes/modern/modern";
```

Error Output:
```
[sass]  error: File to import not found or unreadable: ../../lib/bootstrap-social.
```

The reference is to the first line of _modern.scss
```
@import "../../lib/bootstrap-social";
```

Other Information - 
The bootstrap_social reference in default (which does not give an error) is:
```
@import "node_modules/@ladjs/bootstrap-social/bootstrap-social.scss";
```
Which is also imported in main.scss

**If you change the theme in main.scss and do not force update, it will not update the main.css and you will never see an error or change in styling.**

monkeywithacupcake:
fixes #893

### DIFF
--- a/public/css/themes/modern/_modern.scss
+++ b/public/css/themes/modern/_modern.scss
@@ -1,4 +1,4 @@
-@import "../../lib/bootstrap-social";
+@import "node_modules/@ladjs/bootstrap-social/bootstrap-social.scss";
 
 // Footer
 // -------------------------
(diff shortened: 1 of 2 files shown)


YasharF:
Thank you

PR #908, opened 2018-09-01
fix #854: service error handling

### LINKED ISSUE #854: Last.fm error handling - invalid API key
Last.fm api example crashes with a message [object object] if the API key is invalid.  For example if the API key is changed using:

```
export LASTFM_KEY=fake
export LASTFM_SECRET=is fake
```
A better error handling and error message is desired.

in addition to [object object] being displayed in the browser, the following gets printed out in the console:
```
[object Object]
GET /api/lastfm 500 5169.108 ms - 142
```

jagatfx:
Fix for https://github.com/sahat/hackathon-starter/issues/854

- Added error handler middleware for when `NODE_ENV` is not equal to `development` that sends 500 response with generic `Server Error` message while logging the error to the server console. 
- In lastfm controller, added check for error code from the lastfm service with example of unique handling per code. Adjusted lastfm pug template to display the error message from the service when present and log to the server console. If no code present in error, pass along to generic error handler middleware.

### DIFF
--- a/app.js
+++ b/app.js
@@ -227,6 +227,11 @@ app.get('/auth/pinterest/callback', passport.authorize('pinterest', { failureRed
 if (process.env.NODE_ENV === 'development') {
   // only use in development
   app.use(errorHandler());
+} else {
+  app.use(function(err, req, res, next) {
+    console.error(err);
+    res.status(500).send('Server Error');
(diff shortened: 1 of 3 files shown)


YasharF:
Thank you

PR #910, opened 2018-09-04
Better practice

Adrion:
A clean node installation doesn't require sudo for global dependencies. 
https://docs.npmjs.com/getting-started/fixing-npm-permissions

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ It watches for any changes in your  node.js app and automatically restarts the
 server. Once installed, instead of `node app.js` use `nodemon app.js`. It will
 save you a lot of time in the long run, because you won't need to manually
 restart the server each time you make a small change in code. To install, run
-`sudo npm install -g nodemon`.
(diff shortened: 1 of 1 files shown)


YasharF:
Unfortunately even though a clean node installation shouldn't require sudo from some perspectives, the current default installation of the distribution from nodejs.org tends to do require the sudo permission.   If we remove the `sudo` as part of the command in the readme instructions, many people will end up with errors out of the box which could cause confusion or other slow downs during the hackathons, etc.

Adrion:
It stills a security issue, but facility shouldn't be paid by security risks.

Anyway that's just a Readme statement; In my case I never use `sudo` for npm.
Thanks for the boilerplate though 👍

YasharF:
I don't see a direct threat under OWASP guidelines.  If it is, please contact nodejs.org to make the adjustments in their default installer.

Thank you

PR #918, opened 2018-11-04
Add Helm chart

vymarkov:


YasharF:
From my understanding:
Helm is a tool for managing Kubernetes charts.  There is a gap from going from a hackathon level project to a Kubernetes (Production-Grade Container Orchestration) deployments.  I am not sure if this is a good fit for this project at this point.

vymarkov:
is it no so difficult as it may seem at first glance

PR #919, opened 2018-11-13
Fix wrapping issues in the navbar and logo indentation

estevanmaito:
It collapses the navbar one breakpoint up (md > lg) fixing the alignment of the nav links, but to make this work properly, a Bootstrap 4 bug (https://github.com/twbs/bootstrap/issues/25654) needed to be addressed and this also solved the logo indentation.

As these are visually breaking issues in the current build, I tought it could be pulled separately from the other issues on #853 

Thanks!

### DIFF
--- a/public/css/main.scss
+++ b/public/css/main.scss
@@ -52,7 +52,8 @@ textarea {
 }
 
 .navbar{
-  padding: 0px;
+  padding-top: 0px;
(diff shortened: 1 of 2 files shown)


YasharF:
Thank you!

Alecto:
image
It is written that fixed, but why is this bug present in 4.3.1?

PR #920, opened 2018-11-14
Update README

akupiec:
Deployment to Google Cloud Platform fix
vm: true - option was deprecated and replaced by env: flex


YasharF:
Thank you

PR #921, opened 2018-11-28
Add passport-snapchat and update documentation

nicholasgonzalezsc:
**Background**
Snapchat recently released [Login Kit for web and server flows](https://docs.snapchat.com/docs/login-kit/#web). Given that Snapchat is a popular application with a large user base, it would be helpful to include it in the starter-kit for developers that would like to integrate with it. 

**Changes**
- Integrates `passport-snapchat` which uses Snapchat's Login Kit under the hood
- Updates the documentation to include references to the new package and setup instructions

**Screenshots**

After sign-in, can see the user's name and bitmoji:

screen shot 2018-11-27 at 6 48 30 pm

Basic login button for Snapchat:

screen shot 2018-11-27 at 6 42 19 pm


YasharF:
It this point, I would like input from @sahat for adding any new authentication providers since per https://github.com/sahat/hackathon-starter/pull/57#issuecomment-34229584  

> it's already complicated enough with Local Strategy + OAuth providers; adding yet another way to authenticate users doesn't help to keep this project simple and easy to understand for novices.

Glancing over the code, my other comment would be regarding the changes at https://github.com/sahat/hackathon-starter/pull/921/files#diff-664f851728e6fc6ec664075df484505cL212 .  That section with its flash pop-up seems to be very specific logic for Snapchat.  If we are to add this new complexity, I would rather see it generalized so it could be used for any login kit that does not provide an email address (such as Steam, etc).  But again, I would like some feedback from @sahat before going forward with any of these changes.

sahat:
Snapchat is big enough platform, I am fine with adding it to the project. 

I agree about the oauth unlink being a bit too specific to Snapchat. Twitter doesn’t give out an email either. Perhaps that logic could come simplified and shared with both Snapchat and Twitter? This is a better solution though than the one I initially implemented where it would assign a fake Twitter email address just to get around the email requirement when signing up with Oauth.

nicholasgonzalezsc:
@YasharF @sahat these are good points, I didn't realize the other auth providers also encountered this issue. I'll make the changes to make the unlink functionality more generic and push an update in a little.

YasharF:
Thank you

Issue #915, opened 2018-10-10
Update documentation for use of MongoDB Atlas


YasharF:
Currently the documentation/readme has references to mlab, and they need to be updated for use of MongoDB Atlas instead.

===============
From: mLab
Sent: Tuesday, October 09, 2018 10:28 PM
Subject: mLab is becoming a part of MongoDB, Inc.

Hello from mLab, 
Today we are announcing that mLab is being acquired by MongoDB, Inc. 
....
What this means for you, our customer 
As part of combining the two companies we will be sunsetting the mLab service and migrating all customers to MongoDB Atlas. 

As a customer of mLab you know that we care deeply about your success. You have our commitment that this migration to MongoDB Atlas will be a smooth experience, and that you are migrating onto an excellent service. 

We will be providing you with specific details of the migration plan in forthcoming updates. In the meantime, here is what you should know: 

- You will be given plenty of time to migrate, and nothing will be required of you for at least 4 months. We expect to have all customers migrated to Atlas within the next 12 months. 
- Once migrated, your Atlas database will be hosted on similar hardware and cost the same or less than your original plan on mLab. 
- We will provide tools that allow you to migrate with either minimal downtime or no downtime to your application. 
- All of this applies to users with mLab Heroku add-ons as well. 

Thank you 
...

mgautam98:
I want to work on this.

YasharF:
I made an attempt to use Atlas and ended up with the following error, hence ended up reverting the merge:
```
(node:382) UnhandledPromiseRejectionWarning: MongoError: not authorized on admin to execute command { insert: "system.indexes", documents: [[{ns admin.sessions} {key [{expires 1}]} {name expires_1} {expireAfterSeconds 0} {unique false}]], ordered: true }
    at Function.MongoError.create (/mnt/c/temp/a/hackathon-starter/node_modules/mongodb-core/lib/error.js:31:11)
    at /mnt/c/temp/a/hackathon-starter/node_modules/mongodb-core/lib/connection/pool.js:497:72
    at authenticateStragglers (/mnt/c/temp/a/hackathon-starter/node_modules/mongodb-core/lib/connection/pool.js:443:16)
    at Connection.messageHandler (/mnt/c/temp/a/hackathon-starter/node_modules/mongodb-core/lib/connection/pool.js:477:5)
[...]

PR #925, opened 2018-12-29
Split www server and app, add lambda handler

brennhill:
Makes this closer to a shared-nothing setup which can run serverless as a lambda or use different http servers.

### DIFF
--- a/app.js
+++ b/app.js
@@ -61,8 +61,7 @@ mongoose.connection.on('error', (err) => {
 /**
  * Express configuration.
  */
-app.set('host', process.env.OPENSHIFT_NODEJS_IP || '0.0.0.0');
-app.set('port', process.env.PORT || process.env.OPENSHIFT_NODEJS_PORT || 8080);
+
 app.set('views', path.join(__dirname, 'views'));
 app.set('view engine', 'pug');
 app.use(expressStatusMonitor());
@@ -236,13 +235,5 @@ if (process.env.NODE_ENV === 'development') {
     res.status(500).send('Server Error');
   });
 }
-
-/**
- * Start Express server.
- */
-app.listen(app.get('port'), () => {
-  console.log('%s App is running at http://localhost:%d in %s mode', chalk.green('✓'), app.get('port'), app.get('env'));
-  console.log('  Press CTRL-C to stop\n');
-});
 
 module.exports = app;
--- a/lambdahandler.js
+++ b/lambdahandler.js
@@ -1,0 +1,7 @@
+const awsServerlessExpress = require('aws-serverless-express');
+const app = require('./app');
+
+exports.app = function handleLambdaRequest(event, context, callback) {
+  const server = awsServerlessExpress.createServer(app);
+  awsServerlessExpress.proxy(server, event, context);
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -500,6 +500,15 @@
       "resolved": "https://registry.npmjs.org/asynckit/-/asynckit-0.4.0.tgz",
       "integrity": "sha1-x57Zf380y48robyXkLzDZkdLS3k="
     },
+    "aws-serverless-express": {
+      "version": "3.3.5",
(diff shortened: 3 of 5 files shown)


brennhill:
@YasharF I can put something together, sure.  It depends on which framework you use.  However, in Amazon you can set all the environment variables in the console or in cloudformation.  

As for performance, it's a bit of a mixed bag, because the first request requires a warm up which can take a second or two.  If there are multiple concurrent requests, that will require a similar number of lambda warmups.  That said, for people just trying something out the free aspect is great, as is the use of lambda for a personal website, for instance, to keep it cheap/free.

brennhill:
Actually, I just realized how much has to be done to set this up.  This is either a massive tutorial on amazon or a lesson on the serverless framework.  I think if people are using this on AWS, they need to understand the security concerns and aspects more directly. In order to "plug and play" lambda you need to also setup RDS for mongoDB, etc.

brennhill:
https://www.mongodb.com/blog/post/serverless-development-with-nodejs-aws-lambda-mongodb-atlas

brennhill:
@sahat can you please comment

YasharF:
> https://www.mongodb.com/blog/post/serverless-development-with-nodejs-aws-lambda-mongodb-atlas

I skimmed over this link by the way.  It is nice that there won't be a need for RDS.  We also have https://github.com/sahat/hackathon-starter/issues/915 in the backlog.  I'll spend some time in the next week or so to see if I can come up with a more condensed set of instructions.

YasharF:
I fixed a bunch of issues in the project this weekend including adding instructions for use of MongoDB Atlas (See under https://github.com/sahat/hackathon-starter#1-step-deployment-with-heroku).  But I don't have any more time this weekend to try to see if I can run the project with the changes in this PR as a lambda function.
It would be great if you can take a stab at it to see if it can run as a function, preferably using CLI commands instead of the AWS Console UI.
Some of the other updates also caused the package.json to end up with a conflict.  Would you be able to resolve the conflicts as well?

Thanks

brennhill:
I've fixed package.json.  I'll try and see if I can put together a serverless setup

YasharF:
I'm meeting with @allenmichael , AWS Technical Evangelist, today to talk about options for deployment.

sahat:
I am not sure if it makes sense adding Lambda support for this project, unless it is for learning or experimental purposes. You typically would split your static assets into S3 and your API endpoint functions would move to Lambda. But I guess you can make the same argument about hackathon-starter being hosted on EC2, Heroku, OpenShift with its static assets and API endpoints in the same app.

If you think you can make it a "plug and play" Lambda solution with README instructions then I support this idea. As @YasharF mentioned, it has 1M free requests per month which is a great use case for hackathons and personal projects. I would even add support for DynamoDB which also has several million read/write requests per month for free, unfortunately that would require ditching MongoDB and rewriting more than half of the project.

Or even better would be drop MongoDB in favor of SQLite file bundled with hackathon-starter, which will further simplify the process where you don't even need to start `mongod` process before launching the app. That would also create a smoother transition to Amazon RDS if you so choose to do. But that too would require a massive rewrite of all MongoDB code in the project.

My only reservation for introducing AWS into the project is inherit difficulty associated with setting up your AWS account, setting up user/group/roles in the IAM dashboard, explaining to beginners how critically important it is not to share your AWS access tokens when you upload a project to GitHub, the whole concept behind serverless technology. It could be a bit overwhelming for new developers versus very simple deployment process to Heroku. It is amazing what AWS can do, in fact, I plan to renew my Solution Architect Associate certification and take a AWS Developer Associate certification exam later this year because of how much I like using AWS.

YasharF:
I made a few attempts, but at the end of the day, I was not able to come up with a quick start set of instructions to get the project running on Lambda.  Unless if someone else can come up with the instructions in the next month or so, I'm going to close the PR.  The chances are that by the time AWS has an easy to use way, they may have changed enough that we may have to rework this PR.

Issue #63, opened 2014-02-08
Enable login via browserid/mozilla persona


seafoodbuffet:
Please support a way to login via [mozilla persona](https://login.persona.org) using the [BrowserID](https://developer.mozilla.org/en-US/Persona?redirectlocale=en-US&redirectslug=Persona) protocol. There's a plugin that does this. https://github.com/jaredhanson/passport-browserid

BPopson:
Persona is no longer actively developed by Mozilla. Mozilla has committed to operational and security support of the persona.org services until November 30th, 2016.

On November 30th, 2016, Mozilla will shut down the persona.org services. Persona.org and related domains will be taken offline.

If you run a website that relies on Persona, you need to implement an alternative login solution for your users before this date.

Source: https://developer.mozilla.org/en-US/Persona

PR #930, opened 2019-01-27
Remove Mozilla Persona information from README

GregBrimble:
The [issue referenced](https://github.com/sahat/hackathon-starter/issues/63#issuecomment-178876149) points to [the official deprecation notice of Mozilla Persona](https://developer.mozilla.org/en-US/docs/Archive/Mozilla/Persona).

Since the service was deprecated on November 30th, 2016, no new projects will be integrating it, so I think its safe to remove from the README.


YasharF:
Thank you

PR #933, opened 2019-02-09
Iterate through Github private email addresses on register/login of new user

danielhunt:
`profile._json.email` references the **public** email address of a Github user.

By requesting the `user:email` scope on auth, we can retrieve & iterate over the **private** email addresses on a profile.

`profile.emails` looks like this:
```
[
  { value: 'my@email.com',             primary: false, verified: true },
  { value: 'another@email-domain.com', primary: true,  verified: true },
  { value: 'nospam@spam.com',          primary: false, verified: true }
]
```

See here for info: https://stackoverflow.com/a/37429696

### DIFF
--- a/config/passport.js
+++ b/config/passport.js
@@ -11,6 +11,7 @@ const { Strategy: LinkedInStrategy } = require('passport-linkedin-oauth2');
 const { Strategy: OpenIDStrategy } = require('passport-openid');
 const { OAuthStrategy } = require('passport-oauth');
 const { OAuth2Strategy } = require('passport-oauth');
+const _ = require('lodash');
 
 const User = require('../models/User');
 
@@ -179,7 +180,8 @@ passport.use(new GitHubStrategy({
   clientID: process.env.GITHUB_ID,
   clientSecret: process.env.GITHUB_SECRET,
   callbackURL: '/auth/github/callback',
-  passReqToCallback: true
+  passReqToCallback: true,
+  scope: ['user:email']
 }, (req, accessToken, refreshToken, profile, done) => {
   if (req.user) {
     User.findOne({ github: profile.id }, (err, existingUser) => {
@@ -215,7 +217,7 @@ passport.use(new GitHubStrategy({
           done(err);
         } else {
           const user = new User();
-          user.email = profile._json.email;
+          user.email = _.get(_.orderBy(profile.emails, ['primary', 'verified'], ['desc', 'desc']), [0, 'value'], null);
           user.github = profile.id;
           user.tokens.push({ kind: 'github', accessToken });
           user.profile.name = profile.displayName;


YasharF:
It fails to build.  Also, I suspect that function can be rewritten in a much more compact form using ES8, filters, etc.  I didn't look too closely though.  Can it be compacted into a single line expression without having to use an extra function?

danielhunt:
Thanks for the feedback @YasharF - I've overhauled the function completely, moved it to a util file, and added a heap of tests.
Let me know what you think 👍

sahat:
I would prefer not to have a separate util file just for that. In fact, it can be removed altogether and toTitleCase can be moved to the user controller, the only place where it is being used. I would even add a lodash dependency as it is extremely useful for just about any project and it seems to have something similar called startCase (https://lodash.com/docs/4.17.11#startCase)

sahat:
The Github email logic can be simplified with lodash to a single line:
```js
// grab the first email and retrieve its value
_.get(emails, [0, 'value'])
```
I only have 3 emails on my account, but I think they are all sorted such that primary, verified is always on top, but if I am wrong this can be solved with an additional line before grabbing the first email above:
```js
_.orderBy(emails, ['primary', 'verified'], ['desc', 'desc'])
```

danielhunt:
The separate util file was simply to make testing trivial and clearer - I was torn on that though

On lodash, I hadn't considered that. Do you have a preference for a single line without the fluff for something like this?

sahat:
Got it. Yes, a one liner would be clearer about its intention. Grabbing an object's value from an array doesn't need to be so complex for this use case. I added the latest version of lodash to package.json, if you do a git pull from master you should see it there.

danielhunt:
Thanks for the feedback @sahat - I've pushed a lodash version of the util function for comment. If there's a solid solution, I'll roll it all up into the original file.

This information might be important: the logic I'm trying to enforce here is:

* If there's a primary email address present, use that, else:
* If there's a verified email address present, use that, else:
* If there's a public email address present, use that, else:
* If there's an unverified email address present, use that

I might be making this overly complex - would appreciate your input on that.
Otherwise, if I'm completely honest, [the non-lodash version](https://github.com/sahat/hackathon-starter/commit/309092dc171ef9b5d2541a48d14b16139ded4f3d#diff-513965e3b8d437772efb35044e977aa3) is more readable/obvious to me, than [the lodash one](https://github.com/sahat/hackathon-starter/commit/ad43220435d9cea1322f5a6cef82c6234c496d2e#diff-513965e3b8d437772efb35044e977aa3).

danielhunt:
I'm just now realising how complex I've unintentionally made this.

The whole point of requesting the added `email` scope is to gain access to the 'real' email info for the user.

It should be enough to sort, then pluck the too address and be done with it entirely.
The added work I'm putting in seems almost academic - reflecting a reality that doesn't exist.

I'll peel it back to the bare minimum later on.
It should mean this PR is ultimately a trivial one-line update.

Thanks for the scrutiny on this to both of ye 👍

danielhunt:
Right - new & improved one-liner pushed, and passing the (newly modified) tests.

@sahat if this is closer to your preferred solution, I'll remove the util file and drop this to a single-file-change (I've left it in because the tests are actually super handy)

sahat:
@danielhunt Yeah that looks much better. I'd move that line directly into github passport strategy. The only other util file in that folder I might remove later.

Just one small comment, you do not need to use `|| null`, let it fall back to undefined. Even if let's say you want it to be an empty string, you can give `_.get` a third argument which is used as a default value fallback.

danielhunt:
Done!

Pity about having to drop the unit tests, but the PR is now one line :)

sahat:
Thanks @danielhunt 🎉

PR #934, opened 2019-02-12
Set default db URI in app.js

riazahmed0147:
Updated default MongoDB URI in app.js if environment variable is not set.

### DIFF
--- a/app.js
+++ b/app.js
@@ -51,7 +51,7 @@ const app = express();
 mongoose.set('useFindAndModify', false);
 mongoose.set('useCreateIndex', true);
 mongoose.set('useNewUrlParser', true);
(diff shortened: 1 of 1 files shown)


YasharF:
If you are using this repo as is without modification, you wouldn't need that extra parameter since .env.exampe would cover you.  If you have changed the code or deployed it to an environment which needs its own env variable configuration, then 127.0.0.1 may or may not be valid.

What was the problem that you ran into which this change would fix?

riazahmed0147:
Thanks Yashraf, I revert the code and again checked it worked fine.
I wasn't getting any collection for registered users (I was using telnet and got some topology error in MongoDB compass), so the code is completely fine.

PR #942, opened 2019-03-02
Add Charts Example

T-travis:
This PR is to add an example of chart.js using an API to get closing stock values from Microsoft for the last 7 days


YasharF:
Thank you

YasharF:
By the way, it looks like Alpha Vantage accepts even 'api-key' as the api key right now.  Thanks.

Issue #952, opened 2019-04-08
Use of Axios instead of request.js


YasharF:
requestjs is used at a few places, and they can most likely be replaced with axios.  This issue is to investigate the usage and replace where applicable.

FX-Wood:
Working on it! I aim to do it one file at a time, starting with api.js.

YasharF:
Make sure to manually test the changes.  You may have to obtain the relevant API key(s) for testing.

Thanks

YasharF:
Fixed with https://github.com/sahat/hackathon-starter/pull/954. Thank you @FX-Wood!

PR #954, opened 2019-04-23
Refactored to use axios

FX-Wood:
#### This pr is in reference to #952 - use of Axios instead of request

Updates the routes for:
- [x] New York Times
- [x] Web scraping
- [x] Pinterest

I tried to keep the format of each route as consistent as I could. Each route has been manually tested with corresponding api keys.

Any feedback is welcome!

#### Update:
Done! 
Request is no longer a dependency.

Passport.js:
- [x] steam auth

### DIFF
--- a/config/passport.js
+++ b/config/passport.js
@@ -1,5 +1,5 @@
 const passport = require('passport');
-const request = require('request');
+const axios = require('axios');
 const { Strategy: InstagramStrategy } = require('passport-instagram');
 const { Strategy: LocalStrategy } = require('passport-local');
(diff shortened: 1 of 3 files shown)


FX-Wood:
I'll start working on the remaining references to request, should be done by the end of the week.

FX-Wood:
The steam API route should now handle edge cases like:
- [x] User has a private profile
- [x] User has no recent games
- [x] User's most recent game has no achievements

YasharF:
Thank you and great work!

PR #969, opened 2019-05-23
Update Dockerfile

tanaydin:
Added nodemon to docker as runner for easy development.


YasharF:
Using nodemon is optional; at this point, the preference is not to force it.

PR #970, opened 2019-05-28
Fix for Google Oauth2 profile picture

tanaydin:
Seems Google API changed its response and it is not possible to register a new user with google anymore. This small changes fixes the way to user's profile picture from the JSON data.


YasharF:
Thank you!

PR #972, opened 2019-05-28
Basic list of files and open links from Google Drive.

tanaydin:
A basic list of files and view links from Google Drive.

### DIFF
--- a/app.js
+++ b/app.js
@@ -176,6 +176,7 @@ app.post('/api/upload', upload.single('myFile'), apiController.postFileUpload);
 app.get('/api/pinterest', passportConfig.isAuthenticated, passportConfig.isAuthorized, apiController.getPinterest);
 app.post('/api/pinterest', passportConfig.isAuthenticated, passportConfig.isAuthorized, apiController.postPinterest);
 app.get('/api/google-maps', apiController.getGoogleMaps);
+app.get('/api/google-drive', apiController.getGoogleDrive);
 app.get('/api/chart', apiController.getChart);
 
 /**
@@ -197,7 +198,7 @@ app.get('/auth/github', passport.authenticate('github'));
 app.get('/auth/github/callback', passport.authenticate('github', { failureRedirect: '/login' }), (req, res) => {
   res.redirect(req.session.returnTo || '/');
 });
-app.get('/auth/google', passport.authenticate('google', { scope: 'profile email' }));
+app.get('/auth/google', passport.authenticate('google', { scope: ['profile', 'email', 'https://www.googleapis.com/auth/drive'], accessType: 'offline', prompt: 'consent' }));
 app.get('/auth/google/callback', passport.authenticate('google', { failureRedirect: '/login' }), (req, res) => {
   res.redirect(req.session.returnTo || '/');
 });
--- a/app.js
+++ b/app.js
@@ -176,7 +176,7 @@ app.post('/api/upload', upload.single('myFile'), apiController.postFileUpload);
 app.get('/api/pinterest', passportConfig.isAuthenticated, passportConfig.isAuthorized, apiController.getPinterest);
 app.post('/api/pinterest', passportConfig.isAuthenticated, passportConfig.isAuthorized, apiController.postPinterest);
 app.get('/api/google-maps', apiController.getGoogleMaps);
-app.get('/api/google-drive', apiController.getGoogleDrive);
+app.get('/api/google/drive', passportConfig.isAuthenticated, passportConfig.isAuthorized, apiController.getGoogleDrive);
 app.get('/api/chart', apiController.getChart);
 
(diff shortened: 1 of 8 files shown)


YasharF:
Can you have this example also enable the user to open or download a file that is listed?

views/api/google-drive.pug
```diff
@@ -0,0 +1,13 @@
...
+block content
+  .pb-2.mt-2.mb-4.border-bottom
+    h2
+      i.fab.fa-google.fa-sm
+      | Google Drive API List Of Files
+    for file in files
+      li(title=file.id)
+        =file.name
+
+  script.
```
YasharF: Is this a debug script that needs to be removed?

tanaydin: I've added it because "founding how to debug with pug" a bit confusing... It could be useful, but sure I can remove it.

tanaydin:
I can do. Also I'll care abour checks.

app.js
```diff
@@ -197,7 +198,11 @@ app.get('/auth/github', passport.authenticate('github'));
 app.get('/auth/github/callback', passport.authenticate('github', { failureRedirect: '/login' }), (req, res) => {
   res.redirect(req.session.returnTo || '/');
 });
-app.get('/auth/google', passport.authenticate('google', { scope: 'profile email' }));
+app.get('/auth/google', passport.authenticate('google', { scope: [
+  'profile',
+  'email',
+  'https://www.googleapis.com/auth/drive'
+  ], accessType: 'offline', prompt: 'consent'}));
```
YasharF: I think chaging these lines to a single line `app.get('/auth/google', passport.authenticate('google', { scope: ['profile', 'email', 'https://www.googleapis.com/auth/drive'], accessType: 'offline', prompt: 'consent' }));` is probably going to fix the build error.  You can locally run lint to check by:  `npm run lint`

views/api/google-drive.pug
tanaydin: Removed.

app.js
```diff
@@ -176,6 +176,7 @@ app.post('/api/upload', upload.single('myFile'), apiController.postFileUpload);
 app.get('/api/pinterest', passportConfig.isAuthenticated, passportConfig.isAuthorized, apiController.getPinterest);
 app.post('/api/pinterest', passportConfig.isAuthenticated, passportConfig.isAuthorized, apiController.postPinterest);
 app.get('/api/google-maps', apiController.getGoogleMaps);
+app.get('/api/google-drive', apiController.getGoogleDrive);
```
YasharF: Since accessing the API requires authentication, you would need to check if the user is authenticated and authorized.  Otherwise, if a user that hasn't logged in goes to the /api/google-drive then the server will give a 500 error with 
```
TypeError: Cannot read property 'accessToken' of undefined
    at exports.getGoogleDrive (/mnt/c/temp/g/hackathon-starter/controllers/api.js:714:25)
```
Simply change of the above line to:
`app.get('/api/google-drive', passportConfig.isAuthenticated, passportConfig.isAuthorized, apiController.getGoogleDrive);`

which calls to https://github.com/sahat/hackathon-starter/blob/master/config/passport.js#L590 and https://github.com/sahat/hackathon-starter/blob/master/config/passport.js#L600 to check to see if the user is logged in and if we already have a token for the API and if not to obtain one.

YasharF:
Can you also add a section to readme.md about the config that needs to be added in the Google cloud console for allowing the application to use the google drive APIs?  I tried the code, and it failed with the following error.  It would be nice if there are proactive instructions in the readme so things run smoothly the first time :)

> The API returned an error: Error: Access Not Configured. Drive API has not been used in project xxxxxxxxx before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/drive.googleapis.com/overview?project=xxxxxxxxx then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.

app.js
YasharF: By the way, the Auth path will need to be handled as well since the current isAuthorized will try `/auth/google-drive` which is not defined.  I haven't had a chance to look, but if a separate auth path can be set for the google-drive then perhaps that path can "add" the extra drive permission, while the other google auth path doesn't.  Another option that I can think of would be to rename this example to 'google api example' (/api/google) and have this PR start that example with a google drive example in the view.  In this way, isAuthorize will default to `/auth/google` which is already defined.

tanaydin: What do you think about using 'google' as prefix for URLs like

'/google/drive'
'/google/analytics'

YasharF:
It looks like that PR may have accidentally got closed.  I made some minor edits and merged it through https://github.com/sahat/hackathon-starter/commit/e2871fde95b0871319353448e2fe84b923116ff7

PR #976, opened 2019-06-07
Google oAuth2 refresh token support.

tanaydin:
Hello,

After 1 hour Google tokens are getting expired so, this commit adds refresh token support. 
Solves #977

### DIFF
--- a/controllers/api.js
+++ b/controllers/api.js
@@ -14,6 +14,8 @@ const lob = require('lob')(process.env.LOB_KEY);
 const ig = require('instagram-node').instagram();
 const axios = require('axios');
 const { google } = require('googleapis');
+const refresh = require('passport-oauth2-refresh');
+const User = require('../models/User');
 
 /**
  * GET /api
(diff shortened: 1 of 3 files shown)


controllers/api.js
```diff
@@ -722,7 +724,20 @@ exports.getGoogleDrive = (req, res) => {
...
+          if (err) return console.log(`Requesting new access token failed: ${err}`);
+          User.updateOne({
+            _id: req.user.id,
+            'tokens.accessToken': token.accessToken
+          }, {
+            $set: { 'tokens.$.accessToken': newAccessToken }
+          }).then(() => res.redirect('/api/google/drive')).catch(err => console.log(`User update returned an error: ${err}`));
+        });
+      }
+      return console.log(`The API returned an error: ${err}`);
```
YasharF: The `console.log(`The API returned an error: ${err}`);`  seems to run even if the issue was the need for a refresh token and if the few lines before it was able to successfully refresh the token and render the Google Drive files.

controllers/api.js
```diff
@@ -14,6 +14,8 @@ const lob = require('lob')(process.env.LOB_KEY);
 const ig = require('instagram-node').instagram();
 const axios = require('axios');
 const { google } = require('googleapis');
+const refresh = require('passport-oauth2-refresh');
+const User = require('../models/User');
```
YasharF: Can the refresh-related worked be refactored so it is handled within passport.js config, and not in the controller/api.js?  The controller here gets ripped out of the project during code cleanup, and the refresh function seems to be a core functionality that might be needed anytime that we are dealing with an expiring token.  I suspect if it is refactored that way, then there wouldn't be a need for a dependency on passport-oauth2-refresh or ../models/User here.

I haven't thought about this much, but could the refresh potentially be a part of the isAuthorized method in passport.js config?

tanaydin: I'll think about it.

YasharF:
Closing in favor of https://github.com/sahat/hackathon-starter/commit/dcfad818db297ce4d56412ac491558d185ef1601

PR #1000, opened 2019-08-10
Use app.host variable in app.listen() call

stigok:
I'm assuming this was intended. Or maybe you always want to listen to `0.0.0.0`, but still then it would be nice to log the value of `app.get('host')` when the server starts, IMO.

### DIFF
--- a/app.js
+++ b/app.js
@@ -253,8 +253,8 @@ if (process.env.NODE_ENV === 'development') {
 /**
  * Start Express server.
  */
-app.listen(app.get('port'), () => {
-  console.log('%s App is running at http://localhost:%d in %s mode', chalk.green('✓'), app.get('port'), app.get('env'));
+app.listen(app.get('port'), app.get('host'), () => {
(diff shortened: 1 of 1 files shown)


YasharF:
I am trying to understand the benefit and the use case.  In my dev environment:

Without this patch:
```
$ npm start

> hackathon-starter@5.1.4 start /mnt/c/temp/hackathon-starter
> node app.js

✓ App is running at http://0.0.0.0:8080 in development mode
  Press CTRL-C to stop
```

With this patch:
```
$ npm start

> hackathon-starter@5.1.4 start /mnt/c/temp/hackathon-starter
> node app.js

✓ App is running at http://localhost:8080 in development mode
  Press CTRL-C to stop
```

stigok:
I totally agree with the log message being garbage now, as it outputs an invalid URL.
It's true that it's listening to `localhost`, but its not the whole story. It's actually listening to all interfaces - your loop-back device (`localhost`) being one of them, hence, a bit misleading. But it might not be important at all in this project.

YasharF:
Going to close this, since it is not a bug fix, and it is not addressing any specific use case.

PR #1003, opened 2019-08-12
Flash message after user logged out

pcsingh:


controllers/user.js
```diff
@@ -57,11 +57,13 @@ exports.postLogin = (req, res, next) => {
  */
 exports.logout = (req, res) => {
   req.logout();
-  req.session.destroy((err) => {
+  /* req.session.destroy((err) => {
```
YasharF: OWASP requirement for logout includes invalidating the user session: https://www.owasp.org/index.php/Logout

pcsingh: Sir, can you explain it. I am not able to understand.

YasharF: See the link provided.

PR #1010, opened 2019-10-22
Sign in with Slack

mahammad8564:
Todo: 
- Update README.md
- Add API Examples


.env.example
```diff
@@ -1,5 +1,4 @@
 BASE_URL=http://localhost:8080
-MONGODB_URI=mongodb://localhost:27017/test
```
YasharF: This line shouldn't be removed.

Procfile
```diff
@@ -0,0 +1 @@
+web: node app.js
```
YasharF: why is a Procfile being added?

README.md
```diff
@@ -199,7 +199,7 @@ The same goes for other providers.
 - Find the **Redirect URLs** section, click the **+ Add** button, and enter `http://localhost:8080/auth/snapchat/callback`
 - Find the **Development Environment** section. Click the **Generate** button next to the *Confidential OAuth2 Client* heading within it.
 - Copy and paste the generated *Private Key* and *OAuth2 Client ID* keys into `.env`
-- **Note:** *OAuth2 Client ID* is **SNAPCHAT_ID**, *Private Key* is **SNAPCHAT_SECRET** in `.env`
+- **Note:** *CLIENT ID FOR CONFIDENTIAL OAUTH2 FLOW* is **SNAPCHAT_ID**, *CLIENT SECRET FOR CONFIDENTIAL OAUTH2 FLOW* is **SNAPCHAT_SECRET** in `.env`
```
YasharF: These changes are for snapchat.  I prefer that they are submitted through a separate PR.

YasharF: We would need instructions on how to obtain credentials for login by slack.

YasharF:
One of the challenges with slack is that each user ends up with many different servers, where each server can have different credentials.  I suspect that most users of slack have login by another method such as Gmail, or a social network provider already, and probably log in by slack may not be useful enough to add to the project.  I haven't had a chance to look at Slack's API documentation and would be interested to know the API use cases that we can add and how they would add value to hackathon starter users.

YasharF:
Can also resolve any issues that come up when running the following locally, and if the Travis Build is failing?
`npm run lint`
`npm run test`

Procfile
mahammad8564: For heroku deployment

.env.example
mahammad8564: opps, by mistake

Procfile
YasharF: It is not required from my experiments.  It is out of the scope of "Sign in with Slack" as well, so let's remove it.

mahammad8564: ok

README.md
mahammad8564: ya sure

mahammad8564:
> Can also resolve any issues that come up when running the following locally, and if the Travis Build is failing?
> `npm run lint`
> `npm run test`

fixed

mahammad8564:
> One of the challenges with slack is that each user ends up with many different servers, where each server can have different credentials. I suspect that most users of slack have login by another method such as Gmail, or a social network provider already, and probably log in by slack may not be useful enough to add to the project. I haven't had a chance to look at Slack's API documentation and would be interested to know the API use cases that we can add and how they would add value to hackathon starter users.

PR #1011, opened 2019-10-26
Add React to Recommended Client-side Libraries

tvalentius:


YasharF:
I feel like in the context of a hackathon, trying to point developers to react docs, and expecting them to ramp up during the weekend, might be too big of a jump.

PR #1015, opened 2019-11-03
Rework themes usage as part of post BS update fixes. See 853

peterblazejewicz:
This is a proposal how to fix a topic described in the #853

- remove the modern theme. It's not a default one, it was written some time ago based on the older version of the BS itself. We do not have design specs for that theme.
- instead of using modern theme, allow users to efficiently pick a custom theme from the Bootswatch, instead of offering just one custom theme (Flatly), let them pick whatever theme they like or find appealing for the project that work on at the moment
- simplify project
- add dedicated doc on how to work with themes, so users could start quickly.

The proposed solution should match the primary focus of the project, which is a quick, tournaround during hackatons in the context of picking a custom theme to use for the project, and later work on this choice to fine tune the product.

~~There is a blocker that should be fixed before this one could be considered for review and the merge (if accepted).
The project has hardcoded design dependency to the default theme by using `btn-raised` custom class, which is not a part of the BS itself or the Bootswatch. This brakes design after switching theme.
I'll create new issue with description of the problem and how to fix this quickly.~~
This is covered by PR: #1018 

Thanks!


YasharF:
My preference is not to tie the project down to Bootswatch, and have things set up in a way that the developers can use themes from different places.  For example, https://www.creative-tim.com/bootstrap-themes/free has many good themes.  What is perhaps missing is some instructions on how to easily swap themes from different sources and to make sure the process works.

devdreamsolution (approved):

peterblazejewicz:
let's resurrect if still applicable (there is a log going on in the meantime in BS)

PR #1017, opened 2019-11-04
Use HTML5 native client form validation. Closes #1013

### LINKED ISSUE #1013: Feature: unify backend and client validation details
In the other project based on this one there was an issue with different or lacking validation on the client templates (view layer) while the content is validated on the backend layer.
Let me explain:
- the backend code validates password to be not empty and with minimum lenght of 8 characters:
https://github.com/sahat/hackathon-starter/blob/master/controllers/user.js#L87
```js
if (!validator.isLength(req.body.password, { min: 8 })) validationErrors.push({ msg: 'Password must be at least 8 characters long' });
```
- the view layer that captures that user input do not specify minimum length with the `minlength` attribute:
https://github.com/sahat/hackathon-starter/blob/master/views/account/signup.pug#L15
```pug
input.form-control(type='password', name='password', id='password', placeholder='Password', autocomplete='new-password', required)
```
Some views have no that validation layer added at all:
https://github.com/sahat/hackathon-starter/blob/master/views/contact.pug
while the server still validates user input:
https://github.com/sahat/hackathon-starter/blob/master/controllers/contact.js#L25-L29

This results in the inconsistent UX for the users and it was missed in the updates to the view layer previously (including my own PRs).

Using built-in validation does not require the project to add additional validation with JavaScript. This would work with just built-in features of modern browsers.

Thanks!

/cc @YasharF

peterblazejewicz:
Wherever that is possible to sync client HTML5 form validation with the
rules used on the server to validate user input, use type, min length
and required when applicable:

- add type=email, whenever email input is expected
- add minlength attribute, whenever input lenght is checked
- add required, whenever input is checked against empty values.

Thanks!

### DIFF
--- a/views/account/profile.pug
+++ b/views/account/profile.pug
@@ -9,7 +9,7 @@ block content
     .form-group.row
       label.col-md-3.col-form-label.font-weight-bold.text-right(for='email') Email
       .col-md-7
-        input.form-control(type='email', name='email', id='email', value=user.email, autocomplete='email')
+        input.form-control(type='email', name='email', id='email', value=user.email, autocomplete='email', required)
       .offset-sm-3.col-md-7.pl-3
         if user.emailVerified
           .text-success.font-italic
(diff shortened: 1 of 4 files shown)


YasharF:
Thank you!

PR #1018, opened 2019-11-04
Rework .btn-raised as BS custom button style. Fixes #1016

### LINKED ISSUE #1016: Rework `btn-raised` dependency in the project
There is a custom version of default BS buttons named `btn-raised`, which is a custom class name and ustom style.

https://github.com/sahat/hackathon-starter/blob/17b529d27c28e49833c946f62ba484e8ce71bb7e/views/home.pug#L12
https://github.com/sahat/hackathon-starter/blob/17b529d27c28e49833c946f62ba484e8ce71bb7e/views/api/tumblr.pug#L21
https://github.com/sahat/hackathon-starter/blob/bf0bc204b35f7f49f79e725f6c8c49dd497d7886/views/api/twilio.pug#L32

The `btn-raised` is declared in the default theme:
https://github.com/sahat/hackathon-starter/blob/17b529d27c28e49833c946f62ba484e8ce71bb7e/public/css/themes/default/_default.scss#L19,
so this is the hidden design dependency, as one cannot just switch to the other theme without consideration on how to tackle design of the `btn-raised` in the current theme. As the class is a custom one, the custom styling is missing from the buttons using that class and only default, basic styling is being applied to the elements with that class.

There is a PR #1015 as the part of the #853, but this custom class problem should be first decided in case of #1015 being approved.

Possible solutions:
- move the `btn-raised` to the `public/css/main.scss`, so it would work as custom project style for specific button type
- reword the class `btn-raised` to the `btn-light`, as the `btn-light` from BS 4.* seems to be similar in the design choices (lighter version version of primary design, without primary accent applied). Then in the default theme just overide `btn-light` with custom style used in the theme. This would left design of the `btn-light` same as in the current version (for the default theme) and would apply custom theme `btn-light` design when custom themes are used.

/cc @YasharF

peterblazejewicz:
This is the proposed way of fix issue with the custom `btn-raised` button variant 
being specific only to  the `default` theme in the project, which impacts rendering of the
content if other theme is being used.

This moves definition of the `.btn-raised` out of the `default` theme
definition into the main style project file, so it can be part of the
project style no matter what theme is applied to the project.
If users want to opt-out from that style, it's as easy now as replacing
the `btn-raised` type with e.g built-in one, like `btn-primary`.
The default state of the button tries to retain as much as possible
representation of the button under the default theme definition, while
also providing BS based variants for other states of the button like
active, focused, disabled, etc.

Rendering before:

Screenshot 2019-11-04 at 20 56 44

Rendering after:
Screenshot 2019-11-04 at 21 08 59

Thanks!

### DIFF
--- a/public/css/main.scss
+++ b/public/css/main.scss
@@ -79,6 +79,21 @@ textarea {
   right: 0;
 }
 
+// button raised definition
+// this button uses gradient for the background
+$btn-raised-color: rgb(51, 51, 51);
+$btn-raised-background: #f8f8f8;
+$btn-raised-border: rgb(204, 204, 204);
+$enable-gradients: true;
+
+.btn-raised {
+  @include button-variant(
+    $btn-raised-background,
+    $btn-raised-border,
+    $btn-raised-color
+  )
+}
+
 /*
 * Solves Bootstrap 4 issue
 * https://github.com/twbs/bootstrap/issues/25654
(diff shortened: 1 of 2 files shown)


YasharF:
Thanks!

PR #1022, opened 2019-11-29
Update GitHub API Keys instructions

noftaly:
Hi,
I added `Click on "OAuth Apps"` in the GitHub API keys instruction, because otherwise, we may click on the "Register new application" under the `Github Apps` nav (the default one), which will be embarrassing ^^


YasharF:
Thank you!

PR #1025, opened 2019-12-21
add nodemon package

getspooky:
nodemon is a tool that helps develop node.js based applications by automatically restarting the node application when file changes


YasharF:
Nodemon should be a global install, not a part of the app.

PR #1028, opened 2019-12-22
feat: add logger file 

getspooky:


YasharF:
Unfortunately, the current preference is to make error messages easy to discover since during hackathon crunch developers typically would like less complexity.

PR #1045, opened 2020-03-19
Added 404 Not Found Middleware

anje123:


YasharF:
1. Can you move this further down after the existing Error Handler?
2. Can you change `Not Found` to `Page Not Found`

Thanks

YasharF:
It would probably be helpful to state what page address was not found.

PR #1085, opened 2020-10-06
Fix: Broken twitch integration

princeyesuraj:
Fixes broken twitch integration by upgrading the version of passport-twitch-new.


YasharF:
This is not solving the error at https://hackathon-starter.walcony.com/api/twitch

PR #1092, opened 2020-10-08
Update package.json

soumyadeeptadas:
added packeges:
shortid: for generating shorted simple unique id when required;
node-fetch: for easy getching in node


YasharF:
Marking as SPAM. Please note that submitting unnecessary PRs are not acceptable for Hacktoberfest and are disqualified. This is also noted on Hacktoberfest's website.

Issue #882, opened 2018-07-18
Code coverage improvements


YasharF:
Can we improve the testing by increasing the code coverage?  The current status is:
```
-------------------------------|----------|----------|----------|----------|-------------------|
File                           |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
-------------------------------|----------|----------|----------|----------|-------------------|
All files                      |    23.43 |     1.66 |     0.49 |     25.3 |                   |
 hackathon-starter             |    83.89 |    18.18 |     6.67 |    83.89 |                   |
  app.js                       |    83.89 |    18.18 |     6.67 |    83.89 |... 13,217,221,229 |
 hackathon-starter/config      |     8.48 |        0 |        0 |      9.4 |                   |
  passport.js                  |     8.48 |        0 |        0 |      9.4 |... 48,549,550,552 |
 hackathon-starter/controllers |    14.49 |        0 |        0 |    15.73 |                   |
  api.js                       |    17.44 |        0 |        0 |    18.91 |... 32,634,635,640 |
  contact.js                   |    15.38 |        0 |        0 |    15.38 |... 60,61,62,64,65 |
  home.js                      |       50 |      100 |        0 |       50 |                 6 |
  user.js                      |    10.15 |        0 |        0 |    11.17 |... 70,372,376,379 |
 hackathon-starter/models      |    32.14 |        0 |        0 |       36 |                   |
  User.js                      |    32.14 |        0 |        0 |       36 |... 59,61,62,64,65 |
-------------------------------|----------|----------|----------|----------|-------------------|
```

sahat:
I will add some unit tests, hopefully sometime this weekend, depending on my photo shoot plans, if not then next week. Although it’s kind of pointless to have a high code coverage since this is a boilerplate and not a production app, it could serve as a guide for writing tests, so it’s worth adding them. 

To be honest, I haven’t done much Node.js testing since starting this project, but I did write thousands of test cases of various complexity using React/Redux/Redux-Saga/Jest/Enzyme apps for the past several years. I may end up refactoring some code though to make it more testable.

YasharF:
[...]

PR #1095, opened 2020-10-10
add nyt GET test

lokiiarora:
In favor of #882 .

### DIFF
--- a/test/app.js
+++ b/test/app.js
@@ -57,6 +57,14 @@ describe('GET /api/lastfm', () => {
   });
 });
 
+describe('GET /api/nyt', () => {
+  it('should return 200 OK', (done) => {
+    request(app)
+      .get('/api/nyt')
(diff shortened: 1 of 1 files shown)


YasharF:
If we merge this test, it will force everyone that is submitting PRs for any part of the project to also go obtain valid credentials for NYT before testing the project.  This can be seen in the failed Travis test 
https://travis-ci.org/github/sahat/hackathon-starter/jobs/734518384

I haven't researched this but perhaps you can emulate the NYT's response?  Maybe Nock could be used?  If you are emulating the NYT's response, then you would probably want two or more test cases: an authorized response, a not authorized response, etc

YasharF:
I'll remove the invalid tag once this PR is in a good state.  Looking forward to reviewing the changes.  Thanks.

devdreamsolution (approved):

PR #1097, opened 2020-10-11
Adding paypal GET test

jenesh:
This PR is adding a test in order to increase test coverage. This PR is related to issue #882 .

### DIFF
--- a/test/app.js
+++ b/test/app.js
@@ -81,6 +81,14 @@ describe('GET /api/scraping', () => {
   });
 });
 
+describe('GET /api/paypal', () => {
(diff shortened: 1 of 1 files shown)


YasharF:
I suspect you may need to do something similar to the comment in https://github.com/sahat/hackathon-starter/pull/1095#issuecomment-707174187 since the test is to test the code (aka not fail if the credentials are invalid).

YasharF:
I'll remove the invalid tag once this PR is in a good state.  Looking forward to reviewing the changes.  Thanks.

devdreamsolution (approved):

PR #1098, opened 2020-10-11
Replaced moment.js with date-fns to reduce bundle size

dimshik100:
[moment.js](https://momentjs.com/) is great but it is huge and not tree shakable.
I replaced it with only the required functions from the smaller [date-fns](https://date-fns.org/) and by that reduced date utility library size from **18.2K** gzipped (moment.min.js) to **7K** gzipped (used `add`, `format` and `isBefore` functions)


devdreamsolution (approved):

simon1tan:
Wouldn't dayjs be better? Should be pretty compatible with moment
https://www.npmjs.com/package/dayjs 
Unpacked Size
623 kB

Issue #1110, opened 2020-10-19
Use API keys for SendGrid


YasharF:
**_Email from Sendgrid_**
We would need to use API Key instead of SMTP username passwords for sending email through Sendgrid.

> We are emailing to inform you of an upcoming requirement to update your authentication method with Twilio SendGrid to API keys exclusively by November 18th, 2020 in order to ensure uninterrupted service and improve the security of your account. Our records show that you have used basic authentication with username and password for one or more of your API requests with one or more users of your SendGrid account in the last 180 days. 

> **Why API keys?**
> This is an effort to enhance security for all of our users. Using your account username and password for authentication is less secure than using an API Key. Unlike your username and password, API Keys are uniquely generated and can be set to limit the access and specify permissions for a given request.

> **What action is required?**
> Follow these steps to identify and replace your authentication method to API Keys and then implement Two-Factor Authentication (2FA) for enhanced security.

> **What happens if no action is taken?**
> On November 18th, 2020 we will no longer accept basic authentication with username and password, and we will be requiring 2FA to login to your account. If you attempt to authenticate your API requests or SMTP configuration with username and password for any of your users after that date, your requests will be rejected. 
> We’d like to thank you in advance for your prompt attention to these requirements. If you’d like to learn more about how you can enhance the security of your account, view this post. If you have any questions or need assistance, please visit our documentation or reach out to our Support team. 

> Thank you,
> The Twilio SendGrid Team

Step 1: Research options

Bhavesh27:
@YasharF Can i pick this up? Can you assign me this.

Bhavesh27:
@YasharF I researched about it. We have few external library available to do so.
1. nodemailer-sendgrid-transport
2. nodemailer-sendgrid

Both library last updated 3 yrs ago. nodemailer-sendgrid-transport is developed by SendGrid Team itself. But lost support in 2016.

I can rewrite the nodemailer usage using any one of these plugins.

YasharF:
I would probably need to review and take care of https://github.com/sahat/hackathon-starter/pull/1105 before you get started on this issue.

kazukilapislazuli:
Can I work on this?

YasharF:
Fixed with https://github.com/sahat/hackathon-starter/commit/afd298e3cf08c0f783fe028a320c6f285a697f61

PR #1105, opened 2020-10-14
Refactored sendMail transporter calls #1102

nesarazui:
Created a new helper function to abstract sendMail functionality.

PR #1115, opened 2020-10-27
Use API keys for SendGrid emails

Stasmo:
Fix for #1110 

Maintains backwards compatibility with username and password auth.

I made this change on my local project before looking into what the community was doing. Looks like there might already be people working on it. Also there's #1105 that would conflict with this PR. If you're desperate to get in a fix before Nov 18 then merging this might be an OK option, otherwise this should probably just be used as a reference for the actual fix.


devdreamsolution (approved):

YasharF:
Thank you.  This was valuable to migrate the refactored version of the code.  Resolving in favor of https://github.com/sahat/hackathon-starter/pull/1127

PR #1119, opened 2020-10-29
Add meta-description

YannAries:
I found out that the meta description was missing. It's an important element as it summarize the page content.


YasharF:
The tags are primarily useful for SEO, and with today's SEO requirements there are a whole lot of other items that need to be added.  The app developer would need to add those based on the application.

Issue #964, opened 2019-05-12
Migrate the Instagram API Integration


YasharF:
Facebook is in the process of shutting down the instagram api endpoint due to privacy issues.  Some functions, more specifically for business accounts may be available under the graph api for instagram.
The docs for the new API platform is located at https://developers.facebook.com/products/instagram

Already deprecated form the instagram apis:

- Follower List - to read the list of followers and followed-by users
- Relationships - to follow and unfollow accounts on a user’s behalf
- Commenting on Public Content - to post and delete comments on a user’s behalf on public media
- Commenting - to post and delete comments on a user’s behalf on owned media
- Public Content - to read any public profile info and media on a user’s behalf
- Likes - to like and unlike media on a user’s behalf
- Subscriptions - to receive notifications when media is posted

Planned for shutdown in early 2020:
- Basic - to read a user’s profile info and media

YasharF:
https://github.com/sahat/hackathon-starter/commit/8e60de3777e2988e77518eeb619f86ad8e7bfcea removed the currently deprecated parts of the Instagram API and fixed scope permission issue.

YasharF:
The new API: https://developers.facebook.com/docs/instagram-basic-display-api/

adilanchian:
Going to take a stab at this and hopefully have something for ya shortly :)

adilanchian:
Looked through the code here, not sure if I will have time to complete this so I will leave some thoughts here:
- `instagram-node` is 5 years outdated (https://www.npmjs.com/package/instagram-node) this needs to be removed
- There is a new flow for sign in for Instagram that is not included in `passport` which means a new authentication setup will be needed (https://developers.facebook.com/docs/instagram-basic-display-api/getting-started)
- I wasn't able to find any node packages that support the new Instagram APIs, so this will have to be re-written as well

It sort of stinks because this project is heavily setup with Passport and 3rd party libraries, but doesn't really seem like Instagram will play nice with the rest of this.

More than happy to discuss if needed, but I do think this will a much larger change than expected :)

YasharF:
If I recall correctly, "sign in with Instagram" is no longer available.  Their API is no longer an authentication solution.  The part that is supposed to be there is the Instagram Basic Display which you found.  With that said, I suspect we can move forward without any Instagram specific libraries:
[...]

PR #1120, opened 2020-11-08
Migrate Instagram API (deprecated) to Instagram Basic Display. Issue #964

T-Stewart:
Hey, so we've had a go at this issue and it feels like we're so close to solving it but we have hit a problem which we can't seem to get past. 
Where the request going out to Instagram contains the sessions "user" details, but the callback request it's missing. 
The error is `TypeError : Cannot read property '_id'` of undefined.
This happens when trying to call `GET /auth/instagram/callback/code={authCode}` (this request comes via ngrok), and I believe that up until this call - the request contains the `user{ _id{ ... } }` properties. Then it is completely stripped away on the way back. 
Any idea on what we're missing?


YasharF:
I got stuck trying to test your code at the auth step.  More specifically it looked like I have to create a new Instagram API app and when I tried to do so it pushed me to go through the verification with video of how the app is used.  I created a video and submitted it, but they rejected the scope request for instagram_graph_user_profile and instagram_graph_user_media.  Can a developer (in the scope of a hackathon) put something together without going through video/screencast submission to Instagram for app permissions?  If so, can you add some instructions in Readme.md about obtaining an API key and credentials?  Thanks!

PR #1152, opened 2021-07-18
Mongodb memory

haraldkubota:
I was using this repo for a test NodeJS app to test GitHub Actions, when I realized that I need MongoDB running during tests. While possible to do, it's a dependency I did not see as a hard requirement, so I removed it.
"npm test" will now use mongodb-memory-server to spin up a in-memory mongodb instance during test (see app.js).
A sample GitHub Actions workflow is included too.

I understand this is all not required for a Hackathon starter repo, but if you do include the CD part of CI/CD via GitHub Actions, adding the ability to do consistent tests will make the CD steps easier.


haraldkubota:
Works for GitLab too. This is more a confirmation that the mongodb-memory-server works on both GitHub as well as GitLab CI/CD's runners. And it does!

haraldkubota:
Not sure how to remove the package-lock.json change. I often wipe it out to get the latest updates but that edits that file completely.

haraldkubota:
Canceling this. I'll make a nicer one.

PR #1153, opened 2021-07-19
Remove the need for live MongoDB for test

haraldkubota:
I remove the need for a running MongoDB when doing "npm test" by using a in-memory-instance of MongoDB when tests run (which use MongoDB). E.g. test/model.js does not need MongoDB.

Tested by running it via GitHub and GitLab CI/CD runners and at my local PC.


YasharF:
Reverting.  mongodb-memory-server has recurring compatibility issues that causes the tests not to run without a failure response.  It looks like because it depends on 3rd party binaries that may or may not be available, and once available it may take many weeks before they get added to a mongodb-memory-server release.

Ref: 
https://github.com/nodkz/mongodb-memory-server/issues/732
https://github.com/nodkz/mongodb-memory-server/issues/772
https://github.com/nodkz/mongodb-memory-server/issues/754
https://github.com/nodkz/mongodb-memory-server/issues/632

YasharF:
Reverted by https://github.com/sahat/hackathon-starter/commit/12fa2a0b81b451096050fc83cedc72c0903f9409

YasharF:
Restored this commit with https://github.com/sahat/hackathon-starter/commit/ce6d8e4135885fa6a4aeabe710f9be64233de855 ; we need it for CI, and will do a MongoDB version override to load a newer binary version for the memory server.

PR #1191, opened 2022-05-09
Using env file to save SMS phone number for Twilio

frankgdeng:
It is better keeping Twilio account's phone number in the env file than 'hard code' in api.js


YasharF:
Instead of putting the number in the configs, you may want to just read it from the API and provide it to the user.  Note that you may need to provide the user with a choice (i.e. a drop down) if the user has multiple phone numbers in their Twilio account.  See https://www.twilio.com/docs/phone-numbers/api/incomingphonenumber-resource for additional documentation.

PR #1197, opened 2022-09-13
Resolve logout requires a callback function

Angra974:
Stop having this error when logout. The logic rests the same.


plakxj:
收到！

PR #1201, opened 2022-09-30
Fix new Heroku Logo image url instead old url

mewxz029:
[x] This new image url can work well.


YasharF:
Heroku has discontinued their free tier and I am currently planning to update the docs with another provider.

PR #1205, opened 2022-10-06
Fix Apple Icon width in Readme

andreidobrinski:
Adds `width="17"` to match `height="17"` to Apple icon image.

Before:
before

After:
after


YasharF:
I can't reproduce the issue; the icons appear to be square in Edge, Firefox, and Chrome.
image

PR #1211, opened 2022-10-21
fix broken badge

lokbiswa:
Updated the broken link to the dependencies badge


YasharF:
The link doesn't work.

Issue #1210, opened 2022-10-13
Secure Env Not Used for ENV Keys


mehulbaid:
In ref to #1127 , Sendgrid keys are exposed publically.

I can fix this.

mehulbaid:
@sahat - Please prioritise this.

renovX:
can i solve this issue? Thank you

mehulbaid:
I already am on it

YasharF:
Can you elaborate more on what you mean by "Sendgrid keys are exposed publically" ?

YasharF:
Going to close this issue due to no response.  The original issue "Sendgrid keys are exposed publically" is invalid.

YasharF:
https://github.com/sahat/hackathon-starter/commit/80af16175880763cfdd0a0ebc44298842a7b0fbf

PR #1216, opened 2022-10-31
refactor(secrets): require ignored .env

theRemix:
- display helpful message for initial projects
- clear out all secrets in .env.example

related to #1210 

## Error message

if the user does not create a .env file, the following message is displayed and the process exits.
```
❗You probably forgot to create a '.env' file.❗
      Copy .env.example to .env
```


YasharF:
The current .env setup is by design.

PR #1217, opened 2022-10-31
feat(nvm): set node/npm version and add .nvmrc

theRemix:
- pin node version to lts/gallium
- update readme with install link to nvm


YasharF:
I would rather not adding one more step for users that are trying to get started at this point (install nvm).

PR #1218, opened 2022-10-31
refactor(docker)!: ignores .env

theRemix:
### prevents secrets and environment variables from being baked into the Dockerfile
Secrets and other environment variables should not be baked into an image.
Instead, run the docker container with the `--env-file=` argument.
See [docker run options --env-file](https://docs.docker.com/engine/reference/commandline/run/#options)
Example:
```
docker run -p 8080:8080 --env-file=.env --link=mongo local/hackathon-starter
```

### Dockerfile base image uses matching node version

Node version in `package.json` is 16.
This other PR #1217 pins the lts/gallium node version
Dockerfile base image now uses the same node version

### Dockerfile base image uses smaller alpine os

[alpine base images are much smaller](https://nickjanetakis.com/blog/the-3-biggest-wins-when-using-alpine-as-a-base-docker-image), which also means deploys are faster.

### Dockerfile sets NODE_ENV to production

**❗️Breaking Change**

Sets the default NODE_ENV to production so that dotenv can conditionally skip loading .env

### Conditional requirement of dotenv

dotenv should not be required in production. 
if using docker, use `--env-file=` 
if using docker-compose, use
```yaml
env_file:
  - .env
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,13 @@
-FROM node:12-slim
+FROM node:16.18.0-alpine
+
+ENV NODE_ENV production
+EXPOSE 8080
 
 WORKDIR /starter
-ENV NODE_ENV development
-
-COPY package.json /starter/package.json
+COPY . /starter/
 
 RUN npm install pm2 -g
 RUN npm install --production
 
-COPY .env.example /starter/.env.example
-COPY . /starter
-
 CMD ["pm2-runtime","app.js"]
 
-EXPOSE 8080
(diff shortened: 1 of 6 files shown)


YasharF:
>> prevents secrets and environment variables from being baked into the Dockerfile
The .env.example file is a placeholder. You should not use it for production deployments. dotenv npm module does not use the values if the values are set as part of the environment variables.   See the FAQ section at https://www.npmjs.com/package/dotenv

>> Dockerfile sets NODE_ENV to production
The environment shouldn't be set to production as this is a hackathon project boiler plate.

Feel free to submit a separate PR for the docker version upgrade.  Please use an LTS version, test it and keep the PR limited to the version upgrade.
Perhaps https://github.com/sahat/hackathon-starter/blob/master/prod-checklist.md could use some info on moving from dev to prod for docker based projects.  Feel free to take a stab at it.

PR #1225, opened 2023-06-12
Change pug tags with multiple attributes

soundz77:
Prevents the "You should not have pug tags with multiple attributes" error. 
It occurs due to code such as: 
a.nav-link(class=(title === 'Home') ? 'active' : '')(href='/') Home
on lines 50, 52 and 54 and affects any page that uses header.pug.


YasharF:
Thank you

PR #1249, opened 2023-08-29
Created .nvmrc file with version 20

vimark1:
Allows dev to run `nvm use` to get the correct version of node


YasharF:
The project should work with any currently supported version of NodeJS (aka not in End of Life)  https://github.com/nodejs/release#release-schedule

Devs shouldn't be using End of Life software (CWE-1104: Use of Unmaintained Third-Party Components)

I am not sure why we need to have this file added to specify a node version.  package.json does have a node version specified, but we just set that as the lowest LTS version.

PR #1252, opened 2023-08-29
Updated some minor dependencies after running `npm dedupe`

vimark1:


YasharF:
We recreate the package-lock each time we have a release, so going to skip this PR since it is going to be very short lived and functionally doesn't seem to make much of a change for now.

PR #1271, opened 2023-10-04
change text of Stripe and Foursquare to white

### LINKED ISSUE #1270: Uniform text colour under "API Examples" tab
Right now, the text and the underline of the "Foursquare" and the "Stripe" buttons have different colours, which looks non-uniform. The underline is white, while the text is black. I think it would be better to make the text white as well.
Attaching a screenshot:
image

VatsalBhuva11:
Fixes #1270 
**Change**
Make the text colour of "Stripe" and "Foursquare" buttons to white, in the API Examples tab.

**Reason for change**
Improve the UI by making the colours uniform. Different text and underline colours are not particularly pleasing, unless they have some use-case. I believe that this makes the UI look just a bit better, at least.

**Old UI**
image

**New UI**
image

### DIFF
--- a/views/api/index.pug
+++ b/views/api/index.pug
@@ -19,7 +19,7 @@ block content
             | Facebook
     .col-md-4
       a(href='/api/foursquare', style='color: #fff')
-        .card(style='background-color: #1cafec').mb-3
+        .card(style='background-color: #1cafec').text-white.mb-3
           .card-body
             img(src='https://i.imgur.com/PixH9li.png', height=40, style='padding: 0px 10px 0px 0px')
             | Foursquare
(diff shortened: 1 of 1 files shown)


VatsalBhuva11:
@YasharF kindly have a look at this change. I believe the previous UI did not look very uniform with different colours for the underline and the text.

YasharF:
We don't need this at this point.  The text color is relative to the brightness of the background color; there are other buttons that have black text.

YasharF:
Removed the spam tag as we may have an issue that needs to be addressed, but this PR as is needs to be modified and is invalid.

Issue #1268, opened 2023-10-03
Will make navbar remain on top on scrolling


mohammedasimahmed:
The navbar leaves the view space when I scroll down , I want to fix this , @YasharF could you please assign me this issue

Pratik1603:
Please assign this to me

vaxad:
you can assign this to me!

legendary0911:
can you assign it to me ?

YasharF:
We don't need this at this point

PR #1275, opened 2023-10-05
Made navbar remain on top on scrolling #1268

madabhi:
This PR has made changes to the navbar making it sticky to the top while scrolling. It will remain on top only even after scrolling while the page will scroll properly.
Code: #1268


madabhi:
### This PR has made changes to the navbar making it sticky to the top while scrolling. It will remain on top only even after scrolling while the page will scroll properly.
#### Code: #1268

YasharF:
We don't need this at this point.

PR #1292, opened 2023-11-24
Add Tunnelmole as an open source tunneling option

robbie-cahill:
This PR adds [Tunnelmole](https://github.com/robbie-cahill/tunnelmole-client) as a tunneling option.

Tunnelmole is a FOSS tunneling solution with a growing community. It works out of the box and can be optionally self hosted with the [Tunnelmole Service](https://github.com/robbie-cahill/tunnelmole-service). Both the client and service are open source.

Example:
```
➜  ~ tmole 8000
http://bvdo5f-ip-49-183-170-144.tunnelmole.net is forwarding to localhost:8000
https://bvdo5f-ip-49-183-170-144.tunnelmole.net is forwarding to localhost:8000
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ Features
 - CSRF protection
 - MVC Project Structure
 - Node.js clusters support
-- HTTPS Proxy support (via ngrok, Cloudflare, etc.)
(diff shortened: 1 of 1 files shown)


YasharF:
Thank you, doesn't meet our current needs.

PR #1307, opened 2024-10-03
Fix: Make the prepare script cross-platform using cross-env

prathameshkhandare:
This PR updates the `prepare` script in `package.json` to make it cross-platform using the `cross-env` package. This ensures the script runs properly on both Windows and Unix-based systems.

- Installed `cross-env` as a dev dependency.
- Updated the prepare script to work across different environments.


YasharF:
cross-env is no longer being maintained and its github repo has been archived.  We can't use dependencies that are not actively maintained.

PR #1325, opened 2025-02-17
Downgrade octokit to pre-esm-only version

moikaturns:
Downgrade @octokit/rest to ^20.1.1 because anything from 21.0.0 is a breaking change, throwing runtime error ERR_REQUIRE_ESM (@octokit/rest@21.0.0 and beyond is ESM only and this project is CommonJS). Fix by downgrading @octokit/rest to most recent CommonJS friendly version.


YasharF:
Testing of the project as-is, is not indicating the stated error.

PR #1331, opened 2025-02-23
Non-root deployment

moikaturns:
Tried to deploy the app to a path other than root and ran into some problems with links not working and resources not loading. Had reverse proxy at https://host:port/mysite/ forwarding to https://localhost:8080/ (BASE_URL set to former). Made some changes per this PR to make app work in both root and non-root deployment. 

If there's a better way to achieve please advise.

Used the following to simulate a a non-root deployment pasted into app.js:
```
/**
 * Simulate reverse proxy.
 */
const http = require('http');
const httpProxy = require('http-proxy');
const proxy = httpProxy.createProxyServer({});
http.createServer((req, res) => {
  // Get protocol (http or https)
  const protocol = req.connection.encrypted ? 'https' : 'http';
  // Get host (includes the host and port)
  const host = req.headers.host;
  // Get the full URL (protocol + host + path + query string)
  const fullUrl = `${protocol}://${host}${req.url}`;
  if (fullUrl.indexOf(`${process.env.BASE_URL}`)>=0) {
    req.url = req.url.replace(/^\/mysite/, '');
    proxy.web(req, res, { target: 'http://localhost:8080' }); // Change target as needed
  } else {
    console.log(`failed to load ${fullUrl}`);
    res.writeHead(404, { 'Content-Type': 'text/plain' });
    res.end('404 Not Found');
  }
}).listen(3000, () => console.log('Reverse proxy running on port 3000'));
```
Needs these deps:
```
npm install http --save-dev && npm install http-proxy --save-dev
```

### DIFF
--- a/app.js
+++ b/app.js
@@ -261,6 +261,21 @@ app.use((req, res, next) => {
   res.status(404).send('Page Not Found');
 });
 
+/**
+ * Join base ref to path to form absolute url.
+ *  So when BASE_REF is https://host:port/mysite/
+ *  The path /login becomes https://host:port/mysite/login
+ */
+app.locals.baseRef = path => {
+  function addTrailingSlash(str) {
+    return str.endsWith('/') ? str : str + '/';
+  }
+  function removeLeadingSlash(str) {
+    return str.startsWith('/') ? str.slice(1) : str;
+  }
+  return `${addTrailingSlash(process.env.BASE_URL) + removeLeadingSlash(path)}`;
+};
+
 if (process.env.NODE_ENV === 'development') {
   // only use in development
   app.use(errorHandler());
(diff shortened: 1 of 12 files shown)


YasharF:
Considering that the primary users are hackathon participants.  I am not sure if it would be a good idea to add the extra complexity.  At the moment I am having a hard time seeing the use case in the context of a hackathon.  Feel free to comment on the use case.

moikaturns:
App works well on root as-is so if this is the use case no need for this PR. Maybe worth acknowledging behaviour on README. Am keeping the branch for this change on my fork so if users have this need in future they can pick a commit from there if needs. Thanks again.

PR #1343, opened 2025-04-16
feat: Add trakt.tv API example

YasharF:
- Add trakt.tv API example and a basic limited local image cache to avoid hot-linking trakt.tv hosted images.


YasharF:
I have some additional changes to make to this PR.

Issue #1429, opened 2025-10-06
test(e2e): add End-to-End (E2E) tests for the Web Scraping example


YasharF:
**Prerequisites:**

1. Have Hackathon Starter up and running and be able to load /api/scraping without an error and see a list of links to articles.
2. Be able to run `npm run test:e2e` with all existing tests passing.

**Scope:**
- Create `test/e2e/scraping.e2e.test.js` for E2E testing of /api/scraping with Playwrite. See `test/e2e/github-api.e2e.test.js` as an example.
- All page elements that map to data from the API should be included in the tests.
- `npm run test:e2e` should run all existing and new tests, and they should all pass.

_Please include a screenshot of the console output with your PR submission._

Note 1: You would need to complete the prerequisites before asking for GitHub issue assignment.
Note 2: Please don't just copy-paste AI-generated content (code, issue replies, or PR text), as such submissions will be rejected and may be marked as spam. You can use AI assistants to be more efficient, but the final result should be your work with the help of an assistant, not an assistant's output taken at face value and copy-pasted.

Mrinank-Bhowmick:
I would like to work on this issue and have already completed prerequisites.

Issue #1431, opened 2025-10-06
test(e2e): add End-to-End (E2E) tests for the OpenAI Moderation API integration


YasharF:
**Prerequisites:**

1. You must have obtained a valid OpenAI API Key - see https://github.com/sahat/hackathon-starter?tab=readme-ov-file#obtaining-api-keys
2. Have Hackathon Starter up and running with the OpenAI key and be able to load /ai/openai-moderation and check a sentence that gets flagged and one that does not get flagged.
3. Be able to run `npm run test:e2e` with all existing tests passing.

**Scope:**
- Create `test/e2e/openai-moderation.e2e.test.js` for E2E testing of /ai/openai-moderation with Playwrite. See `test/e2e/github-api.e2e.test.js` as an example.
- The test must submit check at least one sentence that would be flagged and a sentence that would not get flagged.
- All page elements that map to data from the API should be included in the tests.
- `npm run test:e2e` should run all existing and new tests, and they should all pass.

_Please include a screenshot of the console output with your PR submission._

Note 1: You would need to complete the prerequisites before asking for GitHub issue assignment.
Note 2: Please don't just copy-paste AI-generated content (code, issue replies, or PR text), as such submissions will be rejected and may be marked as spam. You can use AI assistants to be more efficient, but the final result should be your work with the help of an assistant, not an assistant's output taken at face value and copy-pasted.

Mrinank-Bhowmick:
I would like to work on this issue and have already started working on it

PR #1435, opened 2025-10-06
Add e2e tests for web scraping

Mrinank-Bhowmick:
Fixes - #1429 
### Changes
Added checks for -
- page basics
- table exists with headers
- num of links must be greater than 25

Screenshot of the console output -
image

Screenshot of webpage - 
image


YasharF:
~~I haven't tested this, but I think it has the same issue as https://github.com/sahat/hackathon-starter/issues/1431 .  I suspect you need to make a change to test/playwright.config.js so tests that have POST would not end up with a mismatch.~~

Resolved

Issue #1425, opened 2025-10-06
test(e2e): add End-to-End (E2E) tests for Lob API integration example


YasharF:
**Prerequisites:**

1. You must have obtained a valid Lob API credentials - https://github.com/sahat/hackathon-starter?tab=readme-ov-file#obtaining-api-keys.
2. Have Hackathon Starter up and running with the Lob API and be able to load /api/lob and download the PDF letter.
3. Be able to run `npm run test:e2e` with all existing tests passing.

**Scope:**
- Create `test/e2e/lob.e2e.test.js` for E2E testing of /api/lob with Playwrite. See `test/e2e/github-api.e2e.test.js` as an example.
- All page elements that map to data from the API should be included in the tests.
- `npm run test:e2e` should run all existing and new tests, and they should all pass.

_Please include a screenshot of the console output with your PR submission._

Note 1: You would need to complete the prerequisites before asking for GitHub issue assignment.
Note 2: Please don't just copy-paste AI-generated content (code, issue replies, or PR text), as such submissions will be rejected and may be marked as spam. You can use AI assistants to be more efficient, but the final result should be your work with the help of an assistant, not an assistant's output taken at face value and copy-pasted.

hemanthsavasere:
Hi @YasharF, I would like to work on this issue, can you please assign this to me

YasharF:
@hemanthsavasere I've assigned this issue to you for the next week or so.  Let me know if you are not able to finish it by then.

YasharF:
I made an adjustments to the scope:

You would need to run  `npx playwright test --config=test/playwright.config.js test/e2e/lob.e2e.test.js` for running your new tests.  You don't need to have all of the e2e tests pass, since the ones in test/e2e/ now all require valid API keys.

YasharF:
Resolved by https://github.com/sahat/hackathon-starter/commit/5ce0de057e78c207c0713aa506ea3b981eb8b858

PR #1441, opened 2025-10-07
test(e2e): add comprehensive Lob API integration tests

hemanthsavasere:
Add end-to-end tests for Lob API integration covering:
- Basic page navigation and error handling
- ZIP code details verification
- USPS letter data validation
- PDF viewer functionality
- External link verification
- npx playwright test --config=test/playwright.config.js test/e2e/lob.e2e.test.js runs the new tests with all passing.

image

### DIFF
--- a/test/e2e/lob.e2e.test.js
+++ b/test/e2e/lob.e2e.test.js
@@ -1,221 +1,114 @@
 const { test, expect } = require('@playwright/test');
 
+/**
+ * Helper function to navigate to Lob API page and check for errors
+ * @param {import('@playwright/test').Page} page - Playwright page object
+ * @throws {Error} If API key is not configured
+ */
+async function navigateToLobPage(page) {
+  await page.goto('/api/lob');
+  await page.waitForLoadState('networkidle');
+
+  // Detect if API returned an error (missing API key, etc.)
+  const errorElement = page.locator('.alert.alert-danger');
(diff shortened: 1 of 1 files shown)


hemanthsavasere:
The Pull requests addresses the https://github.com/sahat/hackathon-starter/issues/1425

hemanthsavasere:
@YasharF  Can you please review the PR. Thanks

YasharF (commented):
Let's just have the test fail if there is no valid API key, so if CI is missing the key, etc. it wouldn't go unnoticed.

At a glance, it looks like some of the test is checking the static text on the page.  You can remove those and just have the e2e test focus on the values that map to the API response, and check the format of what got on the page.  You are doing some of that already like checking the zip code with \d{5}/.  

By the way, you don't need to do a full blown PDF validation; I think you can do something like the following to do a smoke check that we have a PDF:

```
file.size > 0 &&  file.size < 10000000) {  // I suspect we won't get a PDF that is over ~10 MB
file.mimetype !== "application/pdf"
filebuffer.lastIndexOf("%PDF") !== 0
filebuffer.lastIndexOf("%%EOF") < buffer.length - 7
```


Image

hemanthsavasere:
e2e tests are passng after the changes

image

hemanthsavasere:
@YasharF  I have addressed the review comments, can you please review it again.

Issue #1417, opened 2025-10-03
Add Microsoft Login


YasharF:
Pre-requisites: 
- You must have Hackathon Starter up and running with working API keys for GitHub, LinkedIn, and Google, and review the related code on how they work and are different from each other.

Scope:
- Implement Login by Microsoft, without relying on a Microsoft-specific package.
- Add placeholder env keys similar to the existing placeholder's in .example.env
- Add instructions to readme for obtaining the required env key similar to the instructions for the existing providers.

YasharF:
Note:  Within a day of Hacktoberfest, I've seen PRs and posts miss the mark due to overreliance on AI. AI can help, but they often suggests irrelevant or unnecessary changes. Please run the project locally, read the README, review and test, revisit the issue, iterate, and think critically about whether your submission needs revision before opening a PR.

Avenger11764:
Hello @YasharF sir,  please assign me this task. I want to contribute.

IamPriyabrato04:
Hi @YasharF , can I work on this issue?? I have experience in implementing OAuth 2.0 in several projects.

YasharF:
@Avenger11764 @IamPriyabrato04 how far have you each got in the pre-reqs?

AqilJaafree:
Can I contribute to this?
If no one has done it yet?

YasharF:
@AqilJaafree how far have you got with the pre-reqs?

AqilJaafree:
I just logged in with microsoft

YasharF:
This issue is still up for grabs.

Sumant-Reddy:
@YasharF I have set up all the prerequisites, please assign me this as part to hacktober fest 2025.

YasharF:
For this issue, I'm going to accept the first PR that can be merged with almost no edits.

Recently I have had multiple people that either were copy-pasting AI generated code that had issues or stated that they have finished the prerequisites but they had not.

Issues with Help Wanted label are up for grabs: https://github.com/sahat/hackathon-starter/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22help%20wanted%22  I also add new issues every few days.

milan-deori:
i want to work on this project plz assign me

YasharF:
@milan-deori only issues with the Help Wanted label are currently up for grabs: https://github.com/sahat/hackathon-starter/issues?q=state%3Aopen%20label%3A%22help%20wanted%22

PR #1442, opened 2025-10-07
feat: Add Microsoft Login

AqilJaafree:
**Related issues:**
#1417


YasharF:
Some of the changes in this PR don't quite make sense as if they were AI generated content, taken at face value, and copy-pasted.

AqilJaafree:
My mistake here, which part should I redo?

YasharF:
Just review your own diff and go over why each change you made is the correct way to go forward and assess why it may not be correct and how else you could go forward.

Issue #1430, opened 2025-10-06
test(e2e): add End-to-End (E2E) tests for the Together AI/LLM Text Classification


YasharF:
**Prerequisites:**

1. You must have obtained a valid Together.ai API Key - see https://github.com/sahat/hackathon-starter?tab=readme-ov-file#obtaining-api-keys
2. Have Hackathon Starter up and running with the Together.ai key and be able to load /ai/togetherai-classifier and classify "I want a refund" without an error.
3. Be able to run `npm run test:e2e-nokey` with all existing tests passing.

**Scope:**
- Create `test/e2e/togetherai-classifier.e2e.test.js` for E2E testing of /ai/togetherai-classifier with Playwright. See `test/e2e-nokey/github-api.e2e.test.js` as an example.
- The test must submit "I want a refund".
- All page elements that map to data from the API should be included in the tests.
~~- `npm run test:e2e` should run all existing and new tests, and they should all pass.~~
- `npx playwright test --config=test/playwright.config.js test/e2e/togetherai-classifier.e2e.test.js` should run the new tests with all passing.

_Please include a screenshot of the console output with your PR submission._

Note 1: You would need to complete the prerequisites before asking for GitHub issue assignment.
Note 2: Please don't just copy-paste AI-generated content (code, issue replies, or PR text), as such submissions will be rejected and may be marked as spam. You can use AI assistants to be more efficient, but the final result should be your work with the help of an assistant, not an assistant's output taken at face value and copy-pasted.

Yasir761:
Hi @YasharF, I’ve completed all the prerequisites—Together.ai API key is set up, /ai/togetherai-classifier loads correctly, and npm run test:e2e passes all existing tests. I’ve reviewed the patterns in github-api.e2e.test.js and I’m confident I can implement a complete E2E test for the Together AI text classification, including submitting “I want a refund” and verifying all mapped page elements. Could you please assign this issue to me?

YasharF:
@Yasir761 I've assigned it to you for the next week or so.  Let me know if you would not be able to finish it by then. 


[...]

PR #1450, opened 2025-10-09
test: add E2E tests for togetherai-classifier endpoint

Mrinank-Bhowmick:
Fixes - #1430 

### Changes
- Added 2 Playwright E2E tests for /ai/togetherai-classifier endpoint
- Tests cover: page load with error handling, and classification with "I want a refund"
- Validates all UI elements: department result, raw model output, and system prompt

Here are some screenshots - 
image

image

image

### DIFF
--- a/test/e2e/togetherai-classifier.e2e.test.js
+++ b/test/e2e/togetherai-classifier.e2e.test.js
@@ -1,6 +1,8 @@
 const { test, expect } = require('@playwright/test');
 
 test.describe('Together AI Classifier Integration', () => {
+  test.describe.configure({ mode: 'serial' });
+
   test('should launch app, navigate to Together AI Classifier page, and handle API response', async ({ page }) => {
     // Navigate to Together AI Classifier page
     await page.goto('/ai/togetherai-classifier');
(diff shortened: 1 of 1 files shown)


test/e2e/togetherai-classifier.e2e.test.js
```diff
@@ -0,0 +1,75 @@
...
+      await expect(submitButton).toContainText('Classify Department');
+    }
+
+    // Common elements on the page
+    await expect(page.locator('.btn-group a[href*="together.ai"]')).toHaveCount(3);
+    await expect(page.locator('text=/Together AI Dashboard/i')).toBeVisible();
+    await expect(page.locator('text=/API Reference/i')).toBeVisible();
+  });
+
+  test('should classify "I want a refund" and display all classification data', async ({ page }) => {
```
YasharF: Let's rename this to `should classify a user request and display all classification data`

Then add a new test case after this for `should classify "I want a refund" as "Returns and Refunds"`

Mrinank-Bhowmick: ok 👍

Mrinank-Bhowmick: Increased timeout to accommodate rate limiting wait in togetherai API free tier

test/e2e/togetherai-classifier.e2e.test.js
```diff
@@ -0,0 +1,98 @@
+const { test, expect } = require('@playwright/test');
+
```
YasharF, line 2: For the wait timer to actually pause at the end of the second test you need to make sure that playwright is not having multiple workers running them at the same time:

Add `test.describe.configure({ mode: 'serial' });` here before the first test.

test/e2e/togetherai-classifier.e2e.test.js
```diff
@@ -0,0 +1,98 @@
...
+    // Basic page checks
+    await expect(page).toHaveTitle(/Together/);
+    await expect(page.locator('h2')).toContainText('Together AI');
+
+    // Detect if API returned an error (missing API key, etc.)
+    const errorElement = page.locator('.alert.alert-danger');
+    const isError = (await errorElement.count()) > 0;
+
+    if (isError) {
```
YasharF: Let's just fail the test if there is an error, including when the API key is missing.

test/e2e/togetherai-classifier.e2e.test.js
Mrinank-Bhowmick: done

YasharF:
Thank you!

Issue #1407, opened 2025-10-01
Migrate Away From express-flash / connect-flash


YasharF:
Hackathon Starter currently depends on express-flash, which itself depends on connect-flash. Both libraries are unmaintained (last updated around 12 years ago). connect-flash has known bugs, and we’ve been patching it with patch-package for months. Continuing to rely on these packages introduces technical debt and fragility. We want to replace them with a custom inline middleware in app.js that provides the same flash messaging functionality without external dependencies.

### **Scope:**

- Remove express-flash from package.json
- Remove connect-flash from the patches directory
- Add a small custom middleware directly in app.js (after express-session) to perform the same function
- Verify Pug templates (layout.pug) continue to render flash messages.
- Tests
- Update README.md

### **Test cases:**

Core Behavior

1. Set a single flash message: calling req.flash('info', 'hello') stores the message in session.
2. Set multiple flash messages: multiple calls before the next request should accumulate messages.
3. Set multiple messages in one call: req.flash('warning', ['msg1', 'msg2']) stores both.
4. Support multiple types: messages of different types (error, info, success) are stored separately.
5. Retrieve and clear messages: reading messages removes them from the session.
6. Retrieve all messages: calling req.flash() with no args returns all messages grouped by type.
7. Return empty array for unknown type: req.flash('unknown') should return [].

Integration Behavior

8. Consume after read: once displayed, messages should not persist into subsequent requests.
9. Expose to views: res.locals.flash should contain messages for rendering in Pug.
10. No messages by default: if nothing is set, res.locals.flash should be empty.
11. Session isolation: messages from one session should not leak into another.

### **Acceptance Criteria:**

- [ ] Flash messages continue to render correctly in the UI with Bootstrap alerts.
- [ ] No related patch-package hacks remain.
- [ ] All above test cases are implemented using Mocha + Chai + Sinonjs (consistent with Hackathon Starter’s test stack).
- [ ] CI passes without reliance on express-flash or connect-flash.
- [ ] Readme.md is updated and has no reference to express-flash or connect-flash

drippypop:
hi, may you please assign this to me? thank you :)

YasharF:
@drippypop what are your thoughts, plan, etc. for implementation?  This is somewhat of a larger issue to tackle, so I am hoping someone experienced to work on it.

sam28u:
can you assign this to me
[...]

PR #1461, opened 2025-10-13
fix: replace express-flash with custom inline middlware

Prasanth-S7:
Fixes: #1407 

### Changes done in this PR
- Removed express-flash dependency and its patch file
- Implemented custom inline flash middleware directly in app.js
- Middleware now handles setting, retrieving, and clearing flash messages via session
- Updated res.render to automatically inject flash messages into res.locals
- Added unit tests to verify message isolation, consumption, and session behavior

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,6 +486,7 @@ The metadata for Open Graph is only set up for the home page (`home.pug`). Updat
 
 | Name                             | Description                                                          |
 | -------------------------------- | -------------------------------------------------------------------- |
+| **config**/flash.js              | The flash middleware (replacement for express-flash)                 |
 | **config**/morgan.js             | Configuration for request logging with morgan.                       |
 | **config**/nodemailer.js         | Configuration and helper function for sending email with nodemailer. |
 | **config**/passport.js           | Passport Local and OAuth strategies, plus login middleware.          |
@@ -795,7 +796,7 @@ You can also buy many beautifully designed _Bootstrap_ themes at various vendors
 ### How do flash messages work in this project?
 
 Flash messages allow you to display a message at the end of the request and access it on the next request and only the next request. For instance, on a failed login attempt, you would display an alert with some error message, but as soon as you refresh that page or visit a different page and come back to the login page, that error message will be gone. It is only displayed once.
-This project uses a custom flash middlware for flash messages. With this you don't have to explicitly send a flash message to every view inside `res.render()`.
+This project uses a middleware for displaying flash messages. You don't have to explicitly send a flash message to every view inside `res.render()`.
 All flash messages are available in your views via `messages` object by default.
 
 Flash messages have a two-step process. You use `req.flash('errors', { msg: 'Error messages goes here' }`
--- a/config/flash.js
+++ b/config/flash.js
@@ -1,6 +1,8 @@
 const { format } = require('util');
 
-// Flash Middleware
+// Flash Middleware as a replacement for express-flash / connect-flash
+// Those packages are unmaintained and have some issues. This is a simple
+// implementation that provides the same functionality.
 exports.flash = (req, res, next) => {
   if (req.flash) return next();
   req.flash = (type, message, ...args) => {


app.js
```diff
@@ -123,7 +123,50 @@ app.use(
 );
 app.use(passport.initialize());
 app.use(passport.session());
-app.use(flash());
+
+/**
+ * Flash Middlware.
+ */
+const flash = (req, res, next) => {
```
YasharF: You can probably make this middleware more compact by using more modern JS features; potentially like the following but with better names and such to make it easier to read:

```
const flash = (req, res, next) => {
  if (req.flash) return next();
  req.flash = (type, msg, ...rest) => {
    const f = (req.session.flash ||= {});
    if (!type) return (req.session.flash = {}, f);
    if (!msg) return f[type] || [], delete f[type];
    const arr = (f[type] ||= []);
    if (rest.length) arr.push(format(msg, ...rest));
    else if (Array.isArray(msg)) return (arr.push(...msg), arr.length);
    else arr.push(msg);
    return arr;
  };
  res.render = ((r) => function (...a) {
    res.locals.messages = req.flash();
    return r.apply(this, a);
  })(res.render);
  next();
};
```

Prasanth-S7: Refactored. Please take a look now

YasharF:
Can you use Git CLI?  Whatever tooling that you are using seems to have a bug or something and is disregarding the protections that the git workflow has to prevent broken builds to be pushed.

YasharF:
Again. it looks like you are not using git CLI that should have prevented the build failure.  I'm going to have to close this PR as invalid if you keep pushing code that fails the build checks.

Prasanth-S7:
The build was failing at the lint-check step since I have used assignments within the return statements. Now I have removed them. Let me check it locally once and then push it properly

YasharF:
The git pre-commit hook is supposed to fix such issues or block the commit.  If the commit is going thru then there is an issue with your git workflow.

Prasanth-S7:
I've found the root cause. The test script that runs after committing was failing due to the glob pattern in the --exclude flag not being properly interpreted on Windows and it didn't exclude the docs-links.test.js file in the `npm run test` step.
`"test": "c8 --temp-directory tmp/coverage mocha --timeout=60000 --exit --exclude 'test/*links.test.js'"` 
Ideally it should be like this to work on all platforms consistently
`"test": "c8 --temp-directory tmp/coverage mocha --timeout=60000 --exit --exclude \"test/*links.test.js\""`
So I have used git commit with --no-verify flag (Extremely sorry for this). That's why it didn't block my commit

Would it be okay if I update the test script and include this fix in the current PR?

YasharF:
Let's create a sperate PR for the exclude pattern bug.  In that PR, can you also update the build.yml to include an os matrix for the "build" job so we don't end up with a similar broken test script later?  See this as an example: https://github.com/eslint-stylistic/eslint-stylistic/blob/main/.github/workflows/ci.yml#L41

Prasanth-S7:
Sure will create a PR for that

YasharF:
Oh, also I have been thinking more about this PR. Considering that this is more of a core function and don't expect devs that use hackathon-starter to touch it, I think may be it is better to get it out of app.js.  May be just put the middleware code in config/flash.js.
If we can reuse `app.use(flash());` in app.js, then the only change in app.js would become replacing `const flash = require('express-flash');`

YasharF:
I fixed the Windows build issue.  There were more changes that I had to make to get it to work: https://github.com/sahat/hackathon-starter/commit/ef30a8c238053c0b869bb639c2cec10c9b7f842a

Can you rebase your changes?

app.js
```diff
@@ -123,7 +123,39 @@ app.use(
...
+    if (!message) {
+      const retrieved = flashMessages[type] || [];
+      delete flashMessages[type];
+      return retrieved;
+    }
+    const arr = (flashMessages[type] ||= []);
+    if (args.length) arr.push(format(message, ...args));
+    else if (Array.isArray(message)) {
+      arr.push(...message);
+      return arr.length;
```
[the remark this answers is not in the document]
YasharF: I think this is intentional, but I would double check.

Prasanth-S7:
Now I have pushed it properly and it also passed all the pre-commit hooks. I have also migrated the inline flash logic to a separate file in the config folder

YasharF:
Thank you!

PR #1654, opened 2026-05-22
fix: remove openai moderation example and all related files

Jorgeotero1998:


YasharF:
Closing the PR as it is missing the required PR submission template.

Also, at a glance it seems to be incomplete.

PR #1682, opened 2026-06-12
chore: re-introduce lint-staged

### LINKED ISSUE #1612: Re-introduce lint-staged
Re-introduced lint-staged to reduce commit friction by reducing the number of files that are being linted.

Rowkash:
## Checklist

- [x] I acknowledge that submissions that include copy-paste of AI-generated content taken at face value (PR text, code, commit message, documentation, etc.) most likely have errors and hence will be rejected entirely and marked as spam or invalid
- [x] I manually tested the change with a running instance, DB, and valid API keys where applicable
- [ ] Added/updated tests if the existing tests do not cover this change
- [x] README or other relevant docs are updated
- [x] `--no-verify` was not used for the commit(s)
- [x] `npm run lint` passed locally without any errors
- [x] `npm test` passed locally without any errors
- [x] `npm run test:e2e:replay` passed locally without any errors
- [x] `npm run test:e2e:custom -- --project=chromium-nokey-live` passed locally without any errors
- [x] PR diff does not include unrelated changes
- [x] PR title follows Conventional Commits — https://www.conventionalcommits.org/en

## Description
- added **lint-staged** dev dependency
-  changed **pre-commit** husky Bash script with lint-staged
- added **.lintstagedrc.json** config for lint-staged

Closes #1612 

## Screenshots of UI changes (browser) and logs/test results (console, terminal, shell, cmd)
screenshot_20260612_113400
screenshot_20260612_113457


.lintstagedrc.json
YasharF, line 1: Instead of a separate file, I think you can add something like this to package.json:

```
  "lint-staged": {
    "**/*.js": [
      "eslint --fix",
      "prettier --write"
    ],
    "**/*.{json,md,yml,yaml}": "prettier --write",
    ".github/**/*.{yml,yaml}": "prettier --write"
  },
```

YasharF (commented):
There is a table with the list of packages in readme.md, which you need to update as well.

.lintstagedrc.json
YasharF: Your `--ignore-unknown` might be a better flag than listing what to include though.  I haven't looked up the related docs, so please check and verify.

YasharF:
Also, for verification, you should experiment with staging poorly formatted files, or files with lint issues and checking to see if they actually get flagged or fixed/reformatted.  At the same time you can try having unstaged changes with lint and formatting issues in the same files and other files in the project to see if that those changes didn't get touched, or block the commit.

Rowkash:
@YasharF 

1. I thought if there are configs for prettier and eslint so we can add also separate config file for "lint-staged" to stay **package.json** cleaner. Should I add config to **package.json** ?
2. We don't need use filter like  "**/*.js". **lint-staged** can understand "*.js"
screenshot_20260614_143628

3. I left logic like it was. "lint" script runs eslint on all ".js" files and then runs prettier foll all files. My config for **lint-staged** do the same logic.
4.  If we need to format all staged files in commit, so better to use "--ignore-unknown" flag. So for example if prettier see file like husky script it will just ignore it.
5. Should I add more screenshots ? I've already checked some cases to work **lint-staged**. 
- lint-staged works on all staged files included in commit ( both of eslint and format )
- if file has some errors but it's not included in stage so **lint-staged** just ignore it

YasharF:
Sorry for the delay.  Thanks for looking into the glob and that we don't need to use filter like "**/.js".  I think the current pending action items list is:

- [ ] Rebase onto latest master
- [ ] Move the lint-staged config into package.json (under a "lint-staged" key) instead of a separate .lintstagedrc.json file, and delete that file.  I prefer not adding additional files to the project at this point.
- [ ] Update the README dependency table to add a row for lint-staged (same table that lists husky, mocha, etc.).
- [ ] Make sure there is a trailing newline to the new .husky/pre-commit file (minor/cosmetic).

Rowkash:
@YasharF 

- [X] Rebase onto latest master
- [X] Move the lint-staged config into package.json (under a "lint-staged" key) instead of a separate .lintstagedrc.json file, and delete that file. I prefer not adding additional files to the project at this point.
- [X] Update the README dependency table to add a row for lint-staged (same table that lists husky, mocha, etc.).
- [X] Make sure there is a trailing newline to the new .husky/pre-commit file (minor/cosmetic).

